PR #1, opened 2022-08-17
Add initial files

mongoKart:
Copied all files over from docs-golang and modified as appropriate.

For an example of the copy action in .github/workflows, see docs-java/.github/workflows

Staging failing until infra is set up.

JIRA: DOCSP-24526


terakilobyte (approved):
LGTM. We'll eventually want to add the check autobuilder action once dop sets up the repo on their end.

PR #3, opened 2022-08-31
[DOCSP-24538] Connection guide > How to connect

mongoKart:
This page is significantly shorter and simpler than it is for other languages; for example, it doesn't get into how to create a connection string for the various types of MongoDB instances, instead linking out to existing docs. Connection options for this driver will be covered on a separate page.

# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-24583

## Self-Review Checklist


source/fundamentals/connection.txt
```diff
@@ -0,0 +1,25 @@
...
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+Overview
+--------
+
+In this section, you'll learn how to use the {+driver-short+} to connect your application to a MongoDB deployment. Click a link in the following list to jump to a topic: 
+
+- :ref:`How to Connect to MongoDB <csharp-connect-to-mongodb>`
```
terakilobyte, line 25: Will this have more links? A list of one item seems bare.

source/fundamentals/connection/connect.txt
```diff
@@ -0,0 +1,36 @@
...
+
+.. tip::
+
+    See :manual:`the MongoDB Manual</reference/connection-string>`: for more information about creating a connection string.
+
+
+
+
+
+
```
terakilobyte, line 36: issue: I think it's important to point other other connection options, such as replica set connection or direct connection. I'm not sure how I feel about removing the image and explanation of the connection string.

source/fundamentals/connection.txt
mongoKart: Yes, it will link to the other pages once those have been added in separate tickets. I left them out for now so as not to introduce broken links.

PR #5, opened 2022-09-06
DOCSP-24539 - C# Connection Options

mongoKart:
The info on this page was pulled from the [Node.js Connection Options page](https://www.mongodb.com/docs/drivers/node/current/fundamentals/connection/connection-options/), the [URI Options spec](https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst), and the [C# driver code](https://github.com/mongodb/mongo-csharp-driver). All values are correct to the best of my knowledge, though it will still get a technical review from the C# team.

# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-24539

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/connection/connection-options.txt
+++ b/source/fundamentals/connection/connection-options.txt
@@ -1,0 +1,225 @@
+.. _csharp-connection-options:
+
+==================
+Connection Options
+==================
+
+This section describes the MongoDB connection and authentication options
+available in the {+driver-short+}. To use these options, include them as parameters in the connection URI when connecting your client to MongoDB.
+
+.. note::
+  
+  You can set any of the following options (and some others not on this list) with a ``MongoClientSettings`` object instead. See :ref:`csharp-mongo-client-settings` for more information.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 1 1 1 1
+
+   * - Name
+     - Accepted Values
+     - Default Value
+     - Description
+
+   * - **appname**
+     - string
+     - ``null``
+     - The app name the driver passes to the server in the client
+       metadata as part of the connection handshake. The server prints this value to the MongoDB logs once the connection is established. The value is also recorded in the slow query logs and profile collections.
+
+   * - **authMechanism**
+     - string
+     - ``null``
+     - The authentication mechanism the driver uses when connecting to the
+       server. If you don't include this option, the driver uses either ``SCRAM-SHA-1`` or ``SCRAM-SHA-256``, depending on the server version. 
+       
+       See
+       :ref:`authentication mechanisms <csharp-authentication-mechanisms>` for available
+       authentication mechanisms.
+
+   * - **authMechanismProperties**
+     - comma-separated ``key:value`` pairs (for example, ``opt1:val1,opt2:val2``)
+     - ``null``
+     - Other authentication options, such as enabling hostname canonicalization for  GSSAPI.
+
+   * - **authSource**
+     - string
+     - ``null``
+     - The database that connections should authenticate against.
+
+   * - **compressors**
+     - comma-separated list of strings (for example, ``snappy,zlib,zstd``)
+     - ``null``
+     - The preferred compression types, in order, for wire-protocol messages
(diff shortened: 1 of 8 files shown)


source/fundamentals/connection/connection-options.txt
```diff
@@ -0,0 +1,225 @@
...
+==================
+Connection Options
+==================
+
+This section describes the MongoDB connection and authentication options
+available in the {+driver-short+}. To use these options, include them as parameters in the connection URI when connecting your client to MongoDB.
+
+.. note::
+  
+  You can set any of the following options (and some others not on this list) with a ``MongoClientSettings`` object instead. See :ref:`csharp-mongo-client-settings` for more information.
```
terakilobyte, line 12: issue: avoid the parenthetical and just be forthright.

source/fundamentals/connection/connection-options.txt
```diff
@@ -0,0 +1,225 @@
...
+     - ``false``
+     - Specifies whether the driver is connecting to a load balancer.
+
+   * - **localThresholdMS**
+     - nonnegative integer
+     - ``15``
+     - The latency window for server eligibility. If a server's round trip takes longer than the fastest server's round-trip time plus this value, the server isn't eligible for selection. Specifying ``0`` selects the fastest eligible server.
+
+   * - **maxIdleTimeMS**
+     - nonnegative integer
```
terakilobyte, line 91: issue: I wonder if it would be more idiomatic to say "unsigned" since C# has that data type.

source/fundamentals/connection/connection-options.txt
```diff
@@ -0,0 +1,225 @@
...
+       
+       You can include this parameter in the connection URI more than once. If you do, the client treats each instance as a separate tag set. The order of the tags in the URI determines the order for read preference.
+
+   * - **replicaSet**
+     - string
+     - ``null``
+     - The name of the replica set to connect to.
+
+   * - **retryReads**
+     - Boolean
```
terakilobyte, line 148: issue: Isn't the keyword `bool` in C#, not `Boolean`? Otherwise, I believe we should capitalize all types for their .NET type.

terakilobyte (changes requested):
A few comments apply throughout regarding capitalization and values. Additionally, in the other connection options pages we've tried to stay action-oriented, i.e. "Specifies the amount of time the driver will wait attempting to connect to a MongoDB server" vs. "The amount of time...".

source/fundamentals/connection/connection-options.txt
mongoKart: Interesting question. An unsigned integer _is_ more idiomatic to C#; on the other hand:

- This page describes options in the context of the connection string, where there's no such thing as a data type.
- "Nonnegative integer" stays in sync with the spec.
- "Unsigned integer" is less intuitive, IMO.

What do you think?

source/fundamentals/connection/connection-options.txt
mongoKart: Yeah, it's `bool` in C# code but "Boolean" in text. This relates back to the previous comment about the importance of using C# data types in this context.

source/fundamentals/connection/connection-options.txt
terakilobyte: I'll buy it as long as we have reasoned it out.

source/fundamentals/connection/connection-options.txt
terakilobyte: For this I think it should be `bool`,  otherwise, I think we have to change other instances to `String`.

mongoKart: I followed the pattern in the spec for leaving them uncapitalized, but made it "Boolean" because that's technically correct. I'd prefer capitalizing every data type in this column, personally

terakilobyte: Then let's capitalize them. The spec is a good starting point but it's on us to make it idiomatic for the language.

terakilobyte (approved):
lgtm

PR #8, opened 2022-09-09
DOCSP-24530 Find Multiple Usage Example

jordan-smith721:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-24529>

## Self-Review Checklist


source/includes/code-examples/FindManyAsync.cs
```diff
@@ -0,0 +1,96 @@
...
+{
+    private static IMongoCollection<Restaurant> _restaurantsCollection;
+    private static string _mongoConnectionString = "<Your MongoDB URI>";
+
+    public static void Main(string[] args)
+    {
+        Setup();
+
+        // Find multiple documents using builders
+        Console.WriteLine("Finding documents with builders...:");
```
mongoKart, line 19: Since one of the recent versions of C#, you can add `using static System.Console;` to the top of the file, and then you don't have to repeat `Console` every time. We shouldn't do this if it makes the examples confusing, though. What do you think?

source/includes/code-examples/FindManyAsync.cs
```diff
@@ -0,0 +1,96 @@
...
+using MongoDB.Bson.Serialization.Conventions;
+using MongoDB.Driver;
+using MongoDB.Driver.Linq;
+
+namespace CSharpExamples.UsageExamples;
+
+public class FindManyAsync
+{
+    private static IMongoCollection<Restaurant> _restaurantsCollection;
+    private static string _mongoConnectionString = "<Your MongoDB URI>";
```
mongoKart, line 12: To simplify testing, we can probably create a file like `app.config` (similarly to the `.envrc` used for CSFLE) that each dev can use to set their own MongoDB URI.

source/usage-examples/findMany.txt
```diff
@@ -0,0 +1,170 @@
...
+         :end-before: end-find-all-sync
+         :language:  csharp
+         :copyable:
+         :dedent:
+
+Expected Result
+~~~~~~~~~~~~~~~
+
+After running the :ref:`full example <find-runnable-example>`, it returns 
+the following results:
```
mongoKart, line 131:
```suggestion
Running the :ref:`full example <find-runnable-example>` returns 
the following results:
```

source/includes/code-examples/FindManySync.cs
```diff
@@ -0,0 +1,97 @@
...
+        FindMultipleRestaurantsLINQSync();
+
+        Console.WriteLine();
+
+        // Find All restaurants
+        Console.WriteLine("Finding all documents...:");
+        FindAllRestaurantsSync();
+    }
+
+    public static void FindMultipleRestaurantsBuilderSync()
```
mongoKart, line 35: If we're going to be making these complete files publicly visible to readers, we should probably follow best style practices. That means all methods besides `Main` should be `private`.

mongoKart (changes requested):
couple small things + questions

source/includes/code-examples/FindManyAsync.cs
jordan-smith721: I didn't know that was an option. I like it, I don't think it makes things more confusing, I'm good with adding it

source/includes/code-examples/FindManyAsync.cs
jordan-smith721: I think this would be fine for the devs to add to their environments for testing, but my concern with adding it to the "full example" code shown in the repo would be that it would add an extra file that the user would need to add on their own if they are copying these examples. It might make things more complex than necessary if they are just trying to run a quick sample.

source/includes/code-examples/FindManySync.cs
```diff
@@ -0,0 +1,97 @@
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Attributes;
+using MongoDB.Bson.Serialization.Conventions;
+using MongoDB.Driver;
+using static System.Console;
+
+namespace CSharpExamples.UsageExamples;
+
+public class FindMany
```
mongoKart, line 9: Extremely picky but the class name should match the file name. I've been using `<methodName>.cs` and `<methodName>Async.cs`.

source/usage-examples/findMany.txt
```diff
@@ -0,0 +1,170 @@
...
+.. TODO: Add links to references once pages are done
+
+To learn more about builders, see <builders page>.
+
+To learn more about LINQ, see <LINQ page>.
+
+API Documentation
+~~~~~~~~~~~~~~~~~
+
+`Find() <{+api-root+}/html/Overload_MongoDB_Driver_IMongoCollectionExtensions_Find.htm>`__
```
mongoKart, line 170: I used bullets here for an unordered list. Preference?

source/usage-examples/findOne.txt
```diff
@@ -12,8 +12,14 @@ Find a Document
    :depth: 2
    :class: singlecol
 
-You can retrieve a single document from a collection by using the ``Find()``
-method.
+You can retrieve a single document from a ``MongoCollection`` by using the
+`Find() <{+api-root+}/html/Overload_MongoDB_Driver_IMongoCollectionExtensions_Find.htm>`__
```
mongoKart, line 16:
```suggestion
`Find() <{+api-root+}/Overload_MongoDB_Driver_IMongoCollectionExtensions_Find.htm>`__
```

mongoKart (approved):
lgtm + small changes/questions

source/usage-examples/findMany.txt
jordan-smith721: Agreed on bullets - changed

source/includes/code-examples/FindManySync.cs
jordan-smith721: Fixed

PR #11, opened 2022-09-13
add index so it will build

schmalliso:
We can't actually build the site if there's no index.txt at the root so we'll know how the TOC works.

Staging: https://docs-mongodborg-staging.corp.mongodb.com/drivers/csharp/docsworker-xlarge/add-index/
Log: https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=6320e1999ef0f10115230251

### DIFF
--- a/source/index.txt
+++ b/source/index.txt
@@ -1,0 +1,11 @@
+=================
+MongoDB C# Driver
+=================
+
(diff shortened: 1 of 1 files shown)


terakilobyte (approved):
lgtm

PR #13, opened 2022-09-14
DOCSP-24534 - UpdateMany usage example

mongoKart:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-24534
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/csharp/docsworker-xlarge/docsp-24534-update-many/usage-examples/updateMany/

## Self-Review Checklist

### DIFF
--- a/source/includes/code-examples/update-many/Restaurant.cs
+++ b/source/includes/code-examples/update-many/Restaurant.cs
@@ -7,18 +7,7 @@ public class Restaurant
 {
     public ObjectId Id { get; set; }
 
-    public string Name { get; set; }
-
-    [BsonElement("restaurant_id")]
-    public string RestaurantId { get; set; }
-
     [BsonElement("cuisine")]
     public string Cuisine { get; set; }
-
-    public object Address { get; set; }
-
-    public string Borough { get; set; }
-
-    public List<object> Grades { get; set; }
 }
 // end-model
--- a/source/includes/code-examples/update-many/Restaurant.cs
+++ b/source/includes/code-examples/update-many/Restaurant.cs
@@ -1,0 +1,24 @@
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Attributes;
+namespace UsageExamples.UpdateMany;
+
+// start-model
+public class Restaurant
+{
+    public ObjectId Id { get; set; }
+
+    public string Name { get; set; }
+
+    [BsonElement("restaurant_id")]
+    public string RestaurantId { get; set; }
+
+    [BsonElement("cuisine")]
+    public string Cuisine { get; set; }
+
+    public object Address { get; set; }
+
+    public string Borough { get; set; }
+
+    public List<object> Grades { get; set; }
+}
+// end-model
--- a/source/includes/code-examples/update-many/UpdateMany.cs
+++ b/source/includes/code-examples/update-many/UpdateMany.cs
@@ -1,0 +1,86 @@
+using MongoDB.Bson.Serialization.Conventions;
+using MongoDB.Driver;
+using static System.Console;
+
+namespace UsageExamples.UpdateMany
+{
+    public class UpdateMany
+    {
+        private static IMongoCollection<Restaurant> _restaurantsCollection;
+        private const string MongoConnectionString = "<Your MongoDB URI>";
+
+        private const string OldCuisine = "Pizza";
+        private const string NewCuisine = "Pasta and breadsticks";
+        private const string CuisineField = "cuisine";
+
+        public static void Main(string[] args)
+        {
+            Setup();
+
+            // Extra space for console readability 
+            WriteLine();
+
+            // Number of restaurants with old cuisine
+            WriteLine($"Restaurants with {CuisineField} \"{OldCuisine}\" found: {FindCountOfRestaurantsWithCuisine(OldCuisine)}");
+
+            // Update many documents synchronously
+            var syncResult = UpdateManyRestaurants();
+            WriteLine($"Restaurants modified by update: {syncResult.ModifiedCount}");
+
+            // Number of restaurants with new cuisine
+            WriteLine($"Restaurants with {CuisineField} \"{NewCuisine}\" found after update: {FindCountOfRestaurantsWithCuisine(NewCuisine)}");
+
+            // Reset sample data
+            Write("Resetting sample data...");
+            ResetSampleData();
+            WriteLine("done.");
+        }
+
+        private static UpdateResult UpdateManyRestaurants()
+        {
+            // start-update-many
+            const string oldValue = "Pizza";
+            const string newValue = "Pasta and breadsticks";
+
+            var filter = Builders<Restaurant>.Filter
+                .Eq("cuisine", oldValue);
+
+            var update = Builders<Restaurant>.Update
+                .Set(restaurant => restaurant.Cuisine, newValue);
+
+            return _restaurantsCollection.UpdateMany(filter, update);
+            // end-update-many
+        }
+
+        private static long FindCountOfRestaurantsWithCuisine(string cuisineValue)
+        {
+            var filter = Builders<Restaurant>.Filter.Eq(CuisineField, cuisineValue);
+            return _restaurantsCollection.Find(filter).CountDocuments();
+        }
+
+        private static void Setup()
+        {
+            // This allows automapping of the camelCase database fields to our models. 
+            var camelCaseConvention = new ConventionPack { new CamelCaseElementNameConvention() };
+            ConventionRegistry.Register("CamelCase", camelCaseConvention, type => true);
+
+            // Establish the connection to MongoDB and get the restaurants database
+            var mongoClient = new MongoClient(MongoConnectionString);
+            var restaurantsDatabase = mongoClient.GetDatabase("sample_restaurants");
+            _restaurantsCollection = restaurantsDatabase.GetCollection<Restaurant>("restaurants");
+        }
+
+        private static void ResetSampleData()
+        {
+            var filter = Builders<Restaurant>.Filter
+                .Eq(CuisineField, NewCuisine);
+
+            var update = Builders<Restaurant>.Update
+                .Set(restaurant => restaurant.Cuisine, OldCuisine);
+
+            _restaurantsCollection.UpdateMany(filter, update);
+        }
+    }
+}
+
+
--- a/source/usage-examples/updateMany.txt
+++ b/source/usage-examples/updateMany.txt
@@ -1,0 +1,101 @@
+.. _csharp-update-many:
+
+=====================
+Update Many Documents
+=====================
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+You can update more than one document at once using the `UpdateMany() <{+api-root+}/M_MongoDB_Driver_IMongoCollection_1_UpdateMany.htm>`__ method on
+a ``MongoCollection`` object. This method requires a **query filter**, which specifies which documents to update, and an **update** statement, which specifies the changes the driver should make to all documents matching the query filter.
+
+.. tip::
+
+   You can pass an instance of `UpdateOptions <{+api-root+}/T_MongoDB_Driver_UpdateOptions.htm>`__ to the ``UpdateMany()`` method in
+   order to customize its behavior.
+
+Examples
+--------
+
+These examples use the following ``Restaurant`` class as a model:
+
+.. literalinclude:: ../includes/code-examples/update-many/Restaurant.cs
+   :start-after: start-model
+   :end-before: end-model
+   :language: csharp
+   :copyable:
+   :dedent:
+
+.. tabs::
+
+   .. tab:: Async
+      :tabid: update-many-async
+
+      Update Many Documents Asynchronously
+      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+      The following example asynchronously updates the ``cuisine`` of all "Pizza" restaurants in the ``restaurants`` collection to "Pasta and breadsticks": 
+
+      .. literalinclude:: ../includes/code-examples/update-many/UpdateManyAsync.cs
+         :start-after: start-update-many-async
+         :end-before: end-update-many-async
+         :language:  csharp
+         :copyable:
+         :dedent:
+
+   .. tab:: Synchronous
+      :tabid: update-many-sync
+
+      Update Many Documents Synchronously
+      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+      The following example synchronously updates the ``cuisine`` of all "Pizza" restaurants in the ``restaurants`` collection to "Pasta and breadsticks":
+
+      .. literalinclude:: ../includes/code-examples/update-many/UpdateMany.cs
+         :start-after: start-update-many
+         :end-before: end-update-many
+         :language:  csharp
+         :copyable:
+         :dedent:
+
+Expected Result
+~~~~~~~~~~~~~~~
+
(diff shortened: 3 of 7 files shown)


source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,101 @@
...
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+You can update more than one document at once using the `UpdateMany() <{+api-root+}/M_MongoDB_Driver_IMongoCollection_1_UpdateMany.htm>`__ method on
```
biniona-mongodb, line 15: Question (possible issue): I thought the convention was to only provide inline links at the end of the page/section? Is there a reason inline links are provided for `UpdateMany`, `UpdateResult`, and `UpdateOptions`?

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,101 @@
...
+a ``MongoCollection`` object. This method requires a **query filter**, which specifies which documents to update, and an **update** statement, which specifies the changes the driver should make to all documents matching the query filter.
+
+.. tip::
+
+   You can pass an instance of `UpdateOptions <{+api-root+}/T_MongoDB_Driver_UpdateOptions.htm>`__ to the ``UpdateMany()`` method in
+   order to customize its behavior.
```
biniona-mongodb, line 21: Suggestion: I think "in order to" does not add value to the sentence

```suggestion
   You can pass an instance of `UpdateOptions <{+api-root+}/T_MongoDB_Driver_UpdateOptions.htm>`__ to the ``UpdateMany()`` method to customize its behavior.
```

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,101 @@
...
+   :copyable:
+   :dedent:
+
+.. tabs::
+
+   .. tab:: Asynchronous
+      :tabid: update-many-async
+
+      Update Many Documents Asynchronously
+      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
biniona-mongodb, line 41: Issue: The titles on the right hand nav do not take you to the expected example.

Suggestion: Remove the titles from the tabs. I think it makes sense to discuss this as 1 example in an async and sync state (I.E change the title from "Examples" to "Example" and then introduce the example outside of the tabs. This is how it is done for [Node](https://www.mongodb.com/docs/drivers/node/current/usage-examples/findOne/) (JS/TS) for example).

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,101 @@
...
+   
+   ``UpdateMany()`` returns an `UpdateResult <{+api-root+}/T_MongoDB_Driver_UpdateResult.htm>`__ object.
+
+More Information
+----------------
+
+.. _update-many-runnable-example:
+
+For fully runnable examples of the ``UpdateMany()`` operations on this page, see the
+`UpdateMany example <{+example+}/update-many/UpdateMany.cs>`__ and `UpdateManyAsync example <{+example+}/update-many/UpdateManyAsync.cs>`__
```
biniona-mongodb, line 90: Issue: missing period

```suggestion
`UpdateMany example <{+example+}/update-many/UpdateMany.cs>`__ and `UpdateManyAsync example <{+example+}/update-many/UpdateManyAsync.cs>`__.
```

source/includes/code-examples/update-many/Restaurant.cs
```diff
@@ -0,0 +1,24 @@
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Attributes;
+namespace UsageExamples.UpdateMany;
+
+// start-model
+public class Restaurant
```
biniona-mongodb, line 6: Question: Can you simplify the restaurant model to only include fields used in the example? Perhaps only the "cuisine" and "_id" value? If possible, I think this will make the page much shorter without loosing explanatory power.

biniona-mongodb (changes requested):
Looks good, like the changes from the example you shared earlier today.

I think some of these comments might apply to the examples as a whole rather than to just this addition. Let me know if you'd like to discuss anything sync!

source/includes/code-examples/update-many/Restaurant.cs
mongoKart: Good idea.

source/includes/code-examples/update-many/Restaurant.cs
```diff
@@ -0,0 +1,24 @@
...
+    [BsonElement("cuisine")]
+    public string Cuisine { get; set; }
+
+    public object Address { get; set; }
+
+    public string Borough { get; set; }
+
+    public List<object> Grades { get; set; }
+}
+// end-model
```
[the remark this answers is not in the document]
mongoKart: Added. Why is that required for .cs files?

source/includes/code-examples/update-many/UpdateMany.cs
```diff
@@ -0,0 +1,86 @@
...
+            Write("Resetting sample data...");
+            ResetSampleData();
+            WriteLine("done.");
+        }
+
+        private static UpdateResult UpdateManyRestaurants()
+        {
+            // start-update-many
+            const string oldValue = "Pizza";
+            const string newValue = "Pasta and breadsticks";
```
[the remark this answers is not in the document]
mongoKart: My reasoning was that if you were looking at just the code, it would show what each method parameter represents. That way, if a reader wanted to adapt this sample for their own purposes, it would be clear more quickly where they needed to make changes. 

For comparison, here's the version without those constants:

```
private static UpdateResult UpdateManyRestaurants()
{
    var filter = Builders<Restaurant>.Filter
        .Eq("cuisine", "Pizza");

    var update = Builders<Restaurant>.Update
        .Set(restaurant => restaurant.Cuisine, "Pasta and breadsticks");

    return _restaurantsCollection.UpdateMany(filter, update);
}
```

Also happy to rope in the team to see what they prefer!

source/usage-examples/updateMany.txt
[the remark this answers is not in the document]
mongoKart: Good point; that confusion never occurred to me. I intentionally added "at once" because it's also true that you can update multiple documents using `UpdateOne()` (you just need to call it multiple times, obviously). Is there a better way to convey that you can do multiple updates with a single method call? 

* You can update multiple documents in a single operation...
* You can update multiple documents with a single method call...
* Etc.

mongoKart: Do you have a link for that convention? I was using [this page](https://www.mongodb.com/docs/meta/style-guide/style/links-and-cross-references/) and didn't see anything there.

I included the links to immediately show the reader where to go for more info.

source/includes/code-examples/update-many/Restaurant.cs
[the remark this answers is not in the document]
biniona-mongodb: Generally [good practice](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline). I'm not aware of a C# guideline one way or the other, but I'd guess unless it says to not include terminating new line you should. Also github gives a little warning character at the end when there is no new line.

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,101 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+You can update more than one document at once using the `UpdateMany() <{+api-root+}/M_MongoDB_Driver_IMongoCollection_1_UpdateMany.htm>`__ method on
+a ``MongoCollection`` object. This method requires a **query filter**, which specifies which documents to update, and an **update** statement, which specifies the changes the driver should make to all documents matching the query filter.
+
+.. tip::
```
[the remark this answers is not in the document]
mongoKart: Added the title for now.

If we omit this box here (which I'm not opposed to), do we need to provide usage examples showing how this option is used? Or is it okay for usage examples to show just the required params and leave it to the Funds page to show optional ones?

source/usage-examples/updateMany.txt
biniona-mongodb: Unfortunately I don't have a link, it's not in the style guide (just a convention arrived at in a team meeting as we all had different linking styles). I think this convention is used in [go usage examples](https://www.mongodb.com/docs/drivers/go/current/usage-examples/findOne/).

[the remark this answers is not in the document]
biniona-mongodb: I think "you can update multiple documents using the..." is sufficient, but "you can update multiple documents with one operation using the" is also alright.

source/includes/code-examples/update-many/UpdateMany.cs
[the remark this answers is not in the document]
biniona-mongodb: I'm not sure I agree that the addition of the variables makes the code more easy to understand or reuse (and thus justifies the increased vertical scroll), but I'm alright resolving the conversation if you feel it improves the page. Just wanted to point it out to be reconsidered.

I don't think it makes sense to loop team in for this. Feel free to resolve this thread and implement whatever you think is best.

source/usage-examples/updateMany.txt
mongoKart: I'm not sure I understand. Present just the async code (as the Node page does)?

Incidentally, shouldn't the Node page have an "On This Page" widget? When it does, won't the JavaScript/TypeScript tabs be broken?

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,101 @@
...
+         :start-after: start-update-many
+         :end-before: end-update-many
+         :language:  csharp
+         :copyable:
+         :dedent:
+
+Expected Result
+~~~~~~~~~~~~~~~
+
+Running either of the :ref:`full examples <update-many-runnable-example>` prints the following results:
```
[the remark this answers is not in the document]
mongoKart: What's tricky is that the "preceding example" performs the update but doesn't actually print that out. What if we created a page section called "Examples," similar to what Node does, that links to the sample .cs files and then has a subsection called "Expected Result"?

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,101 @@
...
+
+.. tabs::
+
+   .. tab:: Asynchronous
+      :tabid: update-many-async
+
+      Update Many Documents Asynchronously
+      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+      The following example asynchronously updates the ``cuisine`` of all "Pizza" restaurants in the ``restaurants`` collection to "Pasta and breadsticks":
```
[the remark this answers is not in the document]
biniona-mongodb: I like your edit a lot!

Only change I'd make is ".. a ``cuisine`` field with a value of "Pasta and breadsticks" (I like being super explicit about types in docs), but I think your suggested edit is good enough to go in as is.

Feel free to ignore this extra edit/resolve this thread at your convenience.

source/usage-examples/updateMany.txt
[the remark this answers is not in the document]
biniona-mongodb: I think usage example can just show minimal required params, and then edge case (customization) stuff can be left for fundamentals pages and API docs.

I think usage examples value focused presentation over being comprehensive.

source/usage-examples/updateMany.txt
biniona-mongodb: We discussed this sync and clarified initial comment. Seemed like we were alright with initial suggestion. Don't see the edit yet, so I'll leave this thread open.

source/usage-examples/updateMany.txt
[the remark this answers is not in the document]
biniona-mongodb: I think what you could work is to move the link to the fully runnable file into the tabbed content, such that the fully runnable code comes before expected output. So the order is:

- Code Snippet
- Link to github
- Output

This way the page can be read top down and make sense. I think this is ordering is used in [go usage examples](https://www.mongodb.com/docs/drivers/go/current/usage-examples/bulkWrite/).

That would also give readers less to think about as they will only see the link to the file they want.

biniona-mongodb (changes requested):
I like the edits that were made! Still a few from the last round to make, but it's looking good.

source/usage-examples/updateMany.txt
mongoKart: The Go page has some inline links; is there a certain kind we're trying to avoid?

source/usage-examples/updateMany.txt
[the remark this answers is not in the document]
mongoKart: I like this a lot.

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,85 @@
...
+   Restaurants modified by update: 1163
+   Restaurants with cuisine "Pasta and breadsticks" found after update: 1163
+   Resetting sample data...done.
+
+More Information
+----------------
+
+.. TODO: Add links to references once pages are done
+
+.. TODO: To learn more about builders, see <builders page>.
```
mongoKart, line 77: changed to TODO for clarity

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,85 @@
...
+
+API Documentation
+-----------------
+
+* `UpdateMany() <{+api-root+}/M_MongoDB_Driver_IMongoCollection_1_UpdateMany.htm>`__
+* `UpdateManyAsync() <{+api-root+}/M_MongoDB_Driver_IMongoCollection_1_UpdateManyAsync.htm>`__
```
mongoKart, line 83: async link was missing

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,85 @@
...
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+You can update more than one document using the `UpdateMany() <{+api-root+}/M_MongoDB_Driver_IMongoCollection_1_UpdateMany.htm>`__ method on
+a collection object.
```
mongoKart, line 16: In a separate review, Nathan suggested changing ``MongoCollection`` to just "collection"

source/usage-examples/updateMany.txt
biniona-mongodb: Yes,  generally we try to put inline links at the end of sections. We do this so as to not give people a fragmented experience while reading something (we felt that readers shouldn't feel the need to disrupt flow of passage to go somewhere else in the main paragraph if it can be avoided. Sometimes it makes sense to add an inline line mid paragraph but often they can be removed.).

So we try to avoid inline links in the beginning or middle of paragraphs. try to save inline links for the end of a section if possible. In this case, it may be best to save the ``UpdateMany()`` link for the API documentation section as going to that page at the beginning may not help them understand the usage example (which is designed to be legible and self contained).

biniona-mongodb (approved):
LGTM + 2 comments. Great work, and thank you for your patience

source/usage-examples/updateMany.txt
mongoKart: Thanks--makes sense. I'll take them out.

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,85 @@
...
+This example uses the following ``Restaurant`` class as a model:
+
+.. literalinclude:: ../includes/code-examples/update-many/Restaurant.cs
+   :start-after: start-model
+   :end-before: end-model
+   :language: csharp
+   :copyable:
+   :dedent:
+
+The following code updates all documents in the ``restaurants`` collection that have a ``cuisine`` field with the value of "Pizza". After the update, these documents will have a ``cuisine`` field with a value of "Pasta and breadsticks":
```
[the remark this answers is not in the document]
mongoKart: edits made

PR #15, opened 2022-09-15
DOCSP 24536 delete one usage ex

rustagir:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-24536
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/csharp/docsworker-xlarge/DOCSP-24536-deleteOneUsageEx/usage-examples/deleteOne/

## Self-Review Checklist

### DIFF
--- a/source/includes/code-examples/DeleteOneSync.cs
+++ b/source/includes/code-examples/DeleteOneSync.cs
@@ -1,0 +1,70 @@
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Attributes;
+using MongoDB.Bson.Serialization.Conventions;
+using MongoDB.Driver;
+using static System.Console;
+
+namespace CSharpExamples.UsageExamples;
+
+public class DeleteOneSync
+{
+    private static IMongoCollection<Restaurant> _restaurantsCollection;
+    private static string _mongoConnectionString = "<Your MongoDB URI>";
+
+    public static void Main(string[] args)
+    {
+        Setup();
+
+        // Delete a document using builders
+        WriteLine("Deleting a document with builders...:");
+        var result = DeleteARestaurantBuilderSync();
+
+        WriteLine($"Deleted documents: {result.DeletedCount}");
+        // Extra space for console readability 
+        WriteLine();
+        
+    }
+
+    private static DeleteResult DeleteARestaurantBuilderSync()
+    {
+        // start-delete-one-builders-sync
+        var filter = Builders<Restaurant>.Filter
+            .Eq("cuisine", "Continental");
+
+        var result = _restaurantsCollection.DeleteOne(filter);
+        return result;
+        // end-delete-one-builders-sync
+    }
+
+    private static void Setup()
+    {
+        // This allows automapping of the camelCase database fields to our models. 
+        var camelCaseConvention = new ConventionPack { new CamelCaseElementNameConvention() };
+        ConventionRegistry.Register("CamelCase", camelCaseConvention, type => true);
+
+        // Establish the connection to MongoDB and get the restaurants database
+        var mongoClient = new MongoClient(_mongoConnectionString);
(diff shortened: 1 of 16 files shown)


source/includes/code-examples/DeleteOneSync.cs
```diff
@@ -0,0 +1,70 @@
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Attributes;
+using MongoDB.Bson.Serialization.Conventions;
+using MongoDB.Driver;
+using static System.Console;
+
+namespace CSharpExamples.UsageExamples;
+
+public class DeleteOneSync
```
jordan-smith721, line 9: Change this class name and the filename to be DeleteOne rather than DeleteOneSync. This will make it match conventions we are using across the other usage examples

source/includes/code-examples/DeleteOneSync.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        Setup();
+
+        // Delete a document using builders
+        WriteLine("Deleting a document with builders...");
+        var result = DeleteARestaurantBuilderSync();
+
+        WriteLine($"Deleted documents: {result.DeletedCount}");
+        // Extra space for console readability 
+        WriteLine();
```
jordan-smith721, line 24: These lines aren't needed since we aren't adding any other output below the last WriteLine

source/includes/code-examples/DeleteOneSync.cs
jordan-smith721, line 9: I would suggest adding a Restore() method to both the sync and async classes, that adds the deleted file back into the collection at the end of the program run. This will make sure that the users sample data doesn't permanently change and makes the apps re-runnable.

source/includes/code-examples/DeleteOneSync.cs
```diff
@@ -0,0 +1,70 @@
...
+        
+    }
+
+    private static DeleteResult DeleteARestaurantBuilderSync()
+    {
+        // start-delete-one-builders-sync
+        var filter = Builders<Restaurant>.Filter
+            .Eq("cuisine", "Continental");
```
jordan-smith721, line 32: Issue: Deleting a document by its cuisine does not ensure that we are deleting a specific document, since many documents have the cuisine "Continental". Since DeleteOne only deletes the first instance that matches the filter, there's no guarantee we know what's being deleted.

Suggestion: Change the filter to delete a specific known document, either by Name or RestaurantId or both of those.

Applies to Async method as well

source/usage-examples/deleteOne.txt
```diff
@@ -0,0 +1,97 @@
...
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+You can delete a document from a ``MongoCollection`` by using the synchronous
+`DeleteOne() <{+api-root+}/Overload_MongoDB_Driver_IMongoCollectionExtensions_DeleteOne.htm>`__
```
jordan-smith721, line 16: We should add a note under the intro paragraph stating that DeleteOne() will delete the first document that it finds that matches the query filter. See the UpdateOne() usage example page for an example of this

source/usage-examples/deleteOne.txt
```diff
@@ -0,0 +1,97 @@
...
+.. code-block:: none
+
+    Deleting a document with builders...
+    Deleted documents: 1
+
+.. tip:: Sample Datasets
+   
+   These examples use the :atlas:`sample datasets </sample-data>` provided by Atlas.
+   The number of documents returned may differ depending on the data in your
+   collection.
```
jordan-smith721, line 77: This tip will not be necessary once the delete method is changed (per comment above)

jordan-smith721 (changes requested):
Great first pass! I've put a couple comments/suggestions below as discussed in our paired review session.

PR #17, opened 2022-09-19
DOCSP-24537 delete many usageex

rustagir:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-24537
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/csharp/docsworker-xlarge/DOCSP-24537-deleteMany-usageex/usage-examples/deleteMany/

## Self-Review Checklist

- [ ] Is this free of any warnings or errors in the RST?

### DIFF
--- a/source/usage-examples/deleteMany.txt
+++ b/source/usage-examples/deleteMany.txt
@@ -1,0 +1,97 @@
+.. _csharp-delete-multiple:
+
+=========================
+Delete Multiple Documents
+=========================
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+You can delete multiple documents from a collection by using the synchronous
+`DeleteMany() <{+api-root+}/Overload_MongoDB_Driver_IMongoCollectionExtensions_DeleteMany.htm>`__
+method, or the asynchronous `DeleteManyAsync() <{+api-root+}/Overload_MongoDB_Driver_IMongoCollectionExtensions_DeleteManyAsync.htm>`__ method. Both methods use a **query filter** which specifies the search criteria.
+
+Examples
+--------
+
+These examples use the following ``Restaurant`` class as a model:
+
+.. literalinclude:: ../includes/code-examples/DeleteMany.cs
+   :start-after: start-model
+   :end-before: end-model
+   :language: csharp
+   :copyable:
+   :dedent:
+
+Delete Multiple Document Using Builders
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. tabs::
+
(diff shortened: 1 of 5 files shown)


rustagir:
since the `Restaurant` class doesnt include the zipcode subfield under Address, I am considering using a different query filter for this example. LMK what your thoughts are

source/usage-examples/deleteMany.txt
```diff
@@ -0,0 +1,97 @@
+.. _csharp-delete-multiple:
```
mongoKart, line 1: This page is following an older format. I worked with Alek last week to rewrite the UpdateMany page, which is now the format we're using for all Usage Example pages. It should be in upstream/master if you pull from there.

mongoKart (changes requested):
nice job! i mentioned in a comment that we're using a slightly different format for these pages now; you can find it in UpdateMany if you pull from upstream/master. you'll also notice that we factored Restaurant to its own C# class to avoid including it in both classes. Let me know if you have any questions!

source/usage-examples/deleteMany.txt
```diff
@@ -0,0 +1,86 @@
...
+
+.. literalinclude:: ../includes/code-examples/DeleteMany.cs
+   :start-after: start-model
+   :end-before: end-model
+   :language: csharp
+   :copyable:
+   :dedent:
+
+The following code deletes all documents in the ``restaurants`` collection that have a 
+``name`` field which includes the word "Green".
```
mongoKart, line 31: trying to make this more precise. lmk if you can make it even better

```suggestion
The following code deletes all documents in the ``restaurants`` collection whose``name`` field has a value that includes the word "Green".
```

mongoKart:
> since the `Restaurant` class doesnt include the zipcode subfield under Address, I am considering using a different query filter for this example. LMK what your thoughts are

Does the missing zipcode field cause something to break? If not, it seems okay to me.

mongoKart (changes requested):
couple of small questions and suggestions!

source/usage-examples/deleteMany.txt
rustagir: I shortened it a bit, but I still think the meaning is very clear. The phrase "field value" is also used elsewhere in our documentation.

source/includes/code-examples/delete-many/Restaurant.cs
```diff
@@ -0,0 +1,23 @@
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Attributes;
+namespace UsageExamples.UpdateMany;
```
mongoKart, line 3:
```suggestion
namespace UsageExamples.DeleteMany;
```

mongoKart (approved):
lgtm + one small change

PR #21, opened 2022-09-28
DOCSP-24557: FAQ page

rustagir:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-24557
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/csharp/docsworker-xlarge/DOCSP-24557-FAQ-page/faq/

## Self-Review Checklist

### DIFF
--- a/source/faq.txt
+++ b/source/faq.txt
@@ -18,8 +18,8 @@ Why Does the Driver Throw a Timeout During Server Selection?
 Each driver operation requires that you choose a healthy server
 satisfying the :manual:`server selection criteria
 </core/read-preference-mechanics>`. If you do not select an appropriate
-server within the `server selection timeout <{+api-root+}/P_MongoDB_Driver_MongoServerSettings_ServerSelectionTimeout.htm>`__, the driver will throw a
-server selection timeout exception. The exception will look similar to the
+server within the `server selection timeout <{+api-root+}/P_MongoDB_Driver_MongoServerSettings_ServerSelectionTimeout.htm>`__, the driver throws a
+server selection timeout exception. The exception looks similar to the
 following:
 
 .. code-block:: none
@@ -55,7 +55,7 @@ The error message consists of multiple parts:
    server description contains an exhaustive description of its current
    state including information about an endpoint, a server version, a
    server type, and its current health state. If the server is not
-   heathy, ``HeartbeatException`` will contain the exception from the
+   heathy, ``HeartbeatException`` contains the exception from the
    last failed heartbeat. Analyzing the ``HeartbeatException`` on each
    cluster node can assist in diagnosing most server selection issues.
    The following heartbeat exceptions are common:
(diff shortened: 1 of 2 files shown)


rustagir:
All of this material has been adapted from [this page](https://mongodb.github.io/mongo-csharp-driver/2.17/frequently_encountered_issues/) from the current documentation.

source/faq.txt
```diff
@@ -0,0 +1,169 @@
...
+that pool is already at ``maxPoolSize``, which is by default 100, then the requesting
+thread will block in a wait queue, which has a default size of ``5 x
+maxPoolSize`` or 500. If the wait queue is also full, then the driver
+throws a ``MongoWaitQueueFullException``. To resolve this issue, try the
+following steps:
+
+1. Tune your indexes. By improving the performance of your queries,
+   you can reduce the time that operations take and reduce the number of
+   concurrent connections needed for your workload.
+#. If you have long-running, analytical queries, you may wish to isolate
```
terakilobyte, line 97:
```suggestion
#. If you have long-running analytical queries, you may wish to isolate
```

source/faq.txt
```diff
@@ -0,0 +1,169 @@
...
+   them to dedicated analytics nodes using ``readPreferenceTags`` or a hidden
```
terakilobyte, line 98: issue: it might be nice for users to be able to follow a link to read preference tag info.

source/faq.txt
```diff
@@ -0,0 +1,169 @@
...
+   secondary.
+#. Increase ``maxPoolSize`` to allow more simultaneous operations to a
+   given cluster node. If your MongoDB cluster does not have sufficient
+   resources to handle the additional connections and simultaneous
+   workload, performance can decrease due to resource contention
+   on the cluster nodes. Adjust this setting only with careful
+   consideration and testing.
```
terakilobyte, line 105: issue: !! Please verify this with James. You've stated potential downsides and the answer is *rarely* to increase ``maxPoolSize``, as this can cause connection storms and exacerbate other issues, but I wonder if we should even include this point.

source/faq.txt
```diff
@@ -0,0 +1,169 @@
...
+#. Increase ``waitQueueMultiple`` to allow more threads/tasks to block
+   waiting for a connection. This is rarely the right solution, and it is
+   more effective to address the concurrency problems in your
+   application.
```
terakilobyte, line 109: issue: Again, I wonder if we should include this since it is rarely the correct solution. I worry we may present options that users will footgun themselves with.

source/faq.txt
```diff
@@ -0,0 +1,169 @@
...
+   
+Why are Certain LINQ or Builder Expressions Unsupported?
+--------------------------------------------------------
+
+Each LINQ or Builder expression must be translated into MongoDB Query
+Language (MQL), which is executed by the server. This is not
```
terakilobyte, line 115: issue: We no longer use the term MongoDB Query Language or MQL. I'll send you the info in another communication channel about this.

terakilobyte (changes requested):
noted a few things

source/faq.txt
JamesKovacs: There are times where you can and should adjust `maxPoolSize`. That would be an article in and of itself. I think it sufficient to mention that you can do this, but with the provisos mentioned.

source/faq.txt
JamesKovacs: Here I think we should be more careful in our recommendations. `waitQueueMultiple` should only be adjusted as a last resort and only then with careful consideration and understanding. Not sure how to convey an increased sense of caution around this setting. If we omit `waitQueueMultiple` entirely, users will still find it and tweak it (much to their detriment in many cases). So I like having it here as an option but with suitable cautionary language and recommendations on others things to fix first.

terakilobyte (approved):
lgtm

PR #29, opened 2022-10-10
DOCSP 24553: Builders Page

rustagir:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-24553
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/csharp/docsworker-xlarge/DOCSP-24553-builders-pg/fundamentals/builders/

## Self-Review Checklist


source/fundamentals/builders.txt
```diff
@@ -0,0 +1,288 @@
...
+using the ``Price`` and ``Stock`` fields:
+
+.. code-block:: csharp
+   :copyable: true
+
+   var builder = Builders<Flower>.Projection;
+   var projection = builder.Expression(f => new { Profit = f.Price * f.Stock });
+
+The result includes only the ``Profit`` field and automatically excludes
+the``Id`` field. This occurs because the driver inspects the expression
```
terakilobyte, line 168:
```suggestion
the ``Id`` field. This occurs because the driver inspects the expression
```

I found this surprising, as the default behavior in other drivers is to never automatically exclude the ``_id`` field. Perhaps this should be an important note.

source/fundamentals/builders.txt
```diff
@@ -0,0 +1,288 @@
...
+--------
+
+In this guide, you can learn about the helper classes, or **builders**, that
+the {+driver-short+} provides to construct various types used in the
+driver. The driver includes types related to the
+specification of filters, updates, projections, sorts, and index keys,
+and you can use builders to aid in the creation of these definitions.
+Using the builder classes, you leverage the power of the
+{+lang-framework+} compiler and your IDE to find errors during
+development and your IDE for discovery and code completion.
```
terakilobyte, line 25: This is a bit confusing, and looking.

I'd like to see a delineated list of the "types" that builders can produce. I also think this paragraph and the following paragraph overlap. Perhaps structure it so the value proposition of builders is clearly stated at the beginning, and then the types the builder produces and how they're used (projection builders build a projection type, but that probably means more to users if stated as they build a projection operation".

source/fundamentals/builders.txt
```diff
@@ -0,0 +1,288 @@
...
+The driver supports using expression trees to render projections. A
+lambda expression contains all the information necessary to form both
+the projection on the server as well as the client-side result and
+requires no further information.
+
+When you define a ``Find()`` projection with the ``Expression()``
+method to create a lambda expression, the
+projection is run client-side. The driver inspects the lambda expression
+to determine which fields are referenced and automatically constructs a
+server-side projection to return only those fields.
```
terakilobyte, line 154: This feels very dense. What are expression trees? Is there a way to simplify this? Is it necessary to include?

terakilobyte (changes requested):
A few things @jordan-smith721 and I noted. Overall this is great, but we feel it's a bit dense in parts. Also, is there an example we can link to (or create) that combines various builders into a single operation, like a find and a projection?

source/fundamentals/builders.txt
rustagir: I broke up the paragraph, deleted some repetitive information, and presented the tasks in a list format for easier comprehension.

source/fundamentals/builders.txt
rustagir: I removed some of the more jargon-heavy material and combined the paragraphs.

terakilobyte (approved):
lgtm

PR #30, opened 2022-10-11
DOCSP-24546 specify query

jordan-smith721:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-24546
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/csharp/docsworker-xlarge/DOCSP-24546-specify-query/fundamentals/crud/read-operations/specify-query/

## Self-Review Checklist

- [x] Is this free of any warnings or errors in the RST? <- Expected errors
- [] Are all the links working?


source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,286 @@
...
+      :language:  csharp
+
+   .. output:: 
+      :language: none
+      :visible:
+
+      { "_id" : 1, "make" : "Fender", "models" : ["Stratocaster", "Telecaster"], "establishedYear" : 1946, "rating" : 9 }
+
+
+To learn more about using builders, see :ref:`csharp-builders`.
```
jordan-smith721, line 91: All links to this page won't work until this page and builders page are in master

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,286 @@
...
+
+   You can also specify queries using LINQ. To learn more about using LINQ, see
+   :ref:`csharp-linq`.
+
+The examples in this guide use the following documents in a collection called
+``guitars``.
+
+.. code-block:: json
+ 
+   { "_id": 1, "make": "Fender", "models": ["Stratocaster", "Telecaster"], "establishedYear": 1946, "rating": 9 }
```
biniona-mongodb, line 40: Suggestion: The length of these documents makes it so that all output blocks require horizontal scroll to view. Would it be possible to make it such that these documents are less long? Maybe something like

```
{"_id:1, "make":"Fender", "colors":["red","blue"], "qty": 10}
```

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,286 @@
...
+
+.. literalinclude:: ../../../includes/fundamentals/code-examples/specify-query/Guitar.cs
+   :language: csharp
+   :copyable:
+   :dedent:
+
+Literal Values
+--------------
+
+Literal value queries return documents with an exact match to your search criteria.
```
biniona-mongodb, line 57: Question/Issue: Is there a reason "search criteria" is used here instead of "query filter"? I think query filter was defined earlier? Should "search criteria" be defined?

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,286 @@
...
+
+The following example specifies a query as a parameter to the ``Find()`` method,
+using the ``Guitar`` class. The query matches all documents where the ``make``
```
biniona-mongodb, line 60: Question/Issue: What does "using the ``Guitar`` class mean precisely here? Could this be stated precisely somewhere on the page?

I think this means decode/encode documents returned from this collection/client to the ``Guitar`` class correct?

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,286 @@
...
+      :language:  csharp
+
+   .. output:: 
+      :language: none
+      :visible:
+
+      { "_id" : 1, "make" : "Fender", "models" : ["Stratocaster", "Telecaster"], "establishedYear" : 1946, "rating" : 9 }
+
+The following example creates a query filter using builders that matches all 
+documents where the ``make`` field equals "Fender".
```
biniona-mongodb, line 76: Issue: I think it's currently ambiguous whether "a query filter" is using builders or "the following example" is using builders. 

```suggestion
The following example uses builders to create a query filter that matches all 
documents where the ``make`` field equals "Fender".
```

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,286 @@
...
+      { "_id" : 1, "make" : "Fender", "models" : ["Stratocaster", "Telecaster"], "establishedYear" : 1946, "rating" : 9 }
+
+
+To learn more about using builders, see :ref:`csharp-builders`.
+
+Comparison
+----------
+
+Comparison operators analyze the value in a document against the specified value
+in your match criteria. Common comparison operators include: ``>`` or ``Gt()``
```
biniona-mongodb, line 97: Question/Issue: Is there a reason "match criteria" is used here? Elsewhere I believe "query filter" or "search criteria" were used.

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,286 @@
...
+      { "_id" : 1, "make" : "Fender", "models" : ["Stratocaster", "Telecaster"], "establishedYear" : 1946, "rating" : 9 }
+
+
+To learn more about using builders, see :ref:`csharp-builders`.
+
+Logical Operators
+-----------------
+
+Logical operators query data using logic applied to the results of field-level
+operators. Common logical operators include ``&&`` or ``And()`` where all operators must be true,
```
biniona-mongodb, line 142: Question/Issue: What is a "field-level operator"? Could this term be defined on the page? I believe this is the only time it is used.

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,286 @@
...
+-----------------
+
+Logical operators query data using logic applied to the results of field-level
+operators. Common logical operators include ``&&`` or ``And()`` where all operators must be true, 
+and ``||`` or ``Or()`` where at least one of the operators must be true.
+
+For a full list of logical operators, see the :manual:`Logical
+Query Operators </reference/operator/query-logical/>` page. 
+
+The following example specifies a query as a parameter to the ``Find()`` method,
```
biniona-mongodb, line 148: Question/Issue: Is this a "query" or a "query filter"? I think by the current definition this should be a "query filter"?

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,286 @@
...
+Array Operators
+---------------
+
+Array operators query data based on the value or quantity of elements in an array
+field. Array operators include ``All()``, ``Any()``, and ``Size()``.
+
+.. note:: 
+
+   The {+driver-long+} translates the ``Any()`` operator to the MongoDB ``$elemMatch``
+   query selector.
```
biniona-mongodb, line 193: Suggestion: Add link to $elemMatch query selector

```suggestion
.. note:: 

   The {+driver-long+} translates the ``Any()`` operator to the MongoDB ``$elemMatch``
   query selector.
   
   To learn more about the ``$elemMatch`` query selector, see ...
```

biniona-mongodb (changes requested):
Looking good, some things to consider

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,286 @@
...
+Overview
+--------
+
+In this guide, you can learn how to specify a query to match a subset
+of documents using the {+driver-long+}.
+
+Most CRUD operations allow you to narrow the set of matched documents
+by passing an expression as an argument to the method, or by specifying
+matching criteria in a **query filter**. Query filters contain one or more query
+operators to determine which documents to include in the resulting set.
```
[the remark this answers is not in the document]
jordan-smith721: - Added CRUD acronym meaning.

- The wording here says "most CRUD operations" since you don't do any matching on documents with create operations. As far as I know all others do let you.

- I changed the first instance of method to say "driver method" to help make the distinction more clear. Let me know if you have other suggestions.

- I restructured the paragraph to better define a query filter in the way that it is now used throughout the page.

- Removed for new definition

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,286 @@
...
+
+.. tip:: Find All Documents
+
+   Passing an empty query filter matches all documents in the collection.
+
+.. tip:: Using LINQ
+
+   You can also specify queries using LINQ. To learn more about using LINQ, see
+   :ref:`csharp-linq`.
```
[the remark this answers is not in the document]
jordan-smith721: Moved the first admonition down to the Literal Values section as I think it actually fits a bit better there.

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,286 @@
...
+field equals "Fender".
+
+.. io-code-block::
+   :copyable:
+
+   .. input:: ../../../includes/fundamentals/code-examples/specify-query/FindEqPOCO.cs
+      :language:  csharp
+
+   .. output:: 
+      :language: none
```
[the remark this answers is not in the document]
jordan-smith721: No particular reason, just didn't think of it. Changed all instances

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,286 @@
...
+Literal value queries return documents with an exact match to your search criteria.
+
+The following example specifies a query as a parameter to the ``Find()`` method,
+using the ``Guitar`` class. The query matches all documents where the ``make`` 
+field equals "Fender".
+
+.. io-code-block::
+   :copyable:
+
+   .. input:: ../../../includes/fundamentals/code-examples/specify-query/FindEqPOCO.cs
```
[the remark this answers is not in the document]
jordan-smith721: I was having issues with it recognizing the absolute path, but it looks like it must have been a vscode issue because it seems to be working now.

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,286 @@
...
+Comparison
+----------
+
+Comparison operators analyze the value in a document against the specified value
+in your match criteria. Common comparison operators include: ``>`` or ``Gt()`` 
+for "greater than", ``<=`` or ``Lte()`` for "less than or equal to", and ``!=``
+or ``Ne()`` for "not equal to".
+
+For a full list of comparison operators, see the :manual:`Comparison
+Query Operators </reference/operator/query-comparison/>` page.
```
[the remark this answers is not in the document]
jordan-smith721: Is this a team convention, or a recent change? I know that we have been using `ref` in place of full links, but I didn't realize that we weren't using `manual` tags anymore? Just looking for clarification since I've been using `manual` tags in most of the pages I've been writing.

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,286 @@
...
+      { "_id" : 2, "make" : "Gibson", "models" : ["Les Paul", "SG", "Explorer"], "establishedYear" : 1902, "rating" : 8 }
+      { "_id" : 3, "make" : "PRS", "models" : ["Silver Sky", "SE", "Custom"], "establishedYear" : 1985, "rating" : 9 }
+      { "_id" : 5, "make" : "Ibanez", "models" : ["RG", "AZ"], "establishedYear" : 1957, "rating" : 7 }
+
+To learn more about using builders, see :ref:`csharp-builders`.
+
+Evaluation Operators
+--------------------
+
+Evaluation operators query data on higher level logic, like regex and text searches.
```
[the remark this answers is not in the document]
jordan-smith721: Reworded using a mix of the description in the link you provided and the one in the [Go docs](https://www.mongodb.com/docs/drivers/go/current/fundamentals/crud/read-operations/query-document/#evaluation)

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,286 @@
...
+of documents using the {+driver-long+}.
+
+Most CRUD operations allow you to narrow the set of matched documents
+by passing an expression as an argument to the method, or by specifying
+matching criteria in a **query filter**. Query filters contain one or more query
+operators to determine which documents to include in the resulting set.
+
+.. tip:: Find All Documents
+
+   Passing an empty query filter matches all documents in the collection.
```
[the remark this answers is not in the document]
jordan-smith721: I moved this to a section that makes it more relevant, changed the wording a bit, and added an example to make it more clear

source/fundamentals/crud/read-operations/specify-query.txt
jordan-smith721: Changing the model to be shorter will make it so some of the examples don't work. For example, the rating field being added on some of the documents allows demonstration of the Exists() method. I could change "models" to "colors" so that it uses shorter words, but I'm not sure the value of having less horizontal scroll is worth rewriting/retesting every example on the page

source/fundamentals/crud/read-operations/specify-query.txt
jordan-smith721: Changed to query filter

source/fundamentals/crud/read-operations/specify-query.txt
jordan-smith721: Updated to be easier to understand and use less "technical jargon"

source/fundamentals/crud/read-operations/specify-query.txt
jordan-smith721: Changed to query filter

source/fundamentals/crud/read-operations/specify-query.txt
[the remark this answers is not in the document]
biniona-mongodb: This a convention used by most of the docs org now I think. We try to use `ref` whenever possible. The reason for this is that we as an org have tried to commit to never removing ref anchors, so they are generally fairly durable (where as we don't commit to changing URLs of pages).

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,331 @@
...
+   .. input:: /includes/fundamentals/code-examples/specify-query/FindGtPOCO.cs
+      :language:  csharp
+
+   .. output::
+      :language: json
+      :visible:
+
+      { "_id" : 4, "make" : "Kiesel", "models" : ["Ares", "Vader", "Solo"], "establishedYear" : 2015, "rating" : null }
+
+The following example uses builders to create a query filter that matches all
```
biniona-mongodb, line 146: Issue: I believe this is the same query as the preceding example. Could the page make it more clear that these are identical? The current phrasing does not explicitly state that they are the same, and leaves the reader to infer this.


Possible Suggestions:

- Update this to "The following example constructs an identical query filter to the preceding example using builders:"

- Use tabs to visually link the two examples in each section.

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,331 @@
...
+      { "_id" : 1, "make" : "Fender", "models" : ["Stratocaster", "Telecaster"], "establishedYear" : 1946, "rating" : 9 }
+
+
+To learn more about using builders, see :ref:`csharp-builders`.
+
+Logical Operators
+-----------------
+
+Logical operators query data using logic applied to the results of two or more
+sets of search criteria. Common logical operators include:
```
biniona-mongodb, line 168: Question/Issue: I believe "query data" should be "match documents" correct? The operator builds a filter rather than actually searches I believe.

Question/Issue: Is there a reason this section uses the term "search criteria"? I don't believe this term is defined or used outside this section.

Issue: [Avoid introducing lists with sentence fragments](https://www.mongodb.com/docs/meta/style-guide/style/lists/#introductory-text)

Issue: Avoid editorial language like "common". I believe we should instead just introduce the list as a subset of operators.

```suggestion
Logical operators are boolean expressions that match documents which evaluate to true. The following is a list of some logical operators:
```

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,331 @@
...
+      :visible:
+
+      { "_id" : 3, "make" : "PRS", "models" : ["Silver Sky", "SE", "Custom"], "establishedYear" : 1985, "rating" : 9 }
+
+To learn more about using builders, see :ref:`csharp-builders`.
+
+Array Operators
+---------------
+
+Array operators query data based on the value or quantity of elements in an array
```
biniona-mongodb, line 228: Question/Issue: Should "query" be "match" here? I believe the user/driver "queries" the database and the operator "matches" values? Applies whenever applicable if accepted.

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,331 @@
...
+field. Array operators include ``All()``, ``Any()``, and ``Size()``.
```
biniona-mongodb, line 229: Issue: list of[ three or more elements ](https://www.mongodb.com/docs/meta/style-guide/writing/write-clear-concise-sentences-paragraphs/#create-short-paragraphs) not in list directive

Suggestion: Break this into a 3x2 table. Add explanations for what each array operator does.

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,331 @@
...
+
+.. note:: 
+
+   The {+driver-long+} translates the ``Any()`` operator to the MongoDB ``$elemMatch``
```
biniona-mongodb, line 233: Question/Issue: I believe the ``Any()`` specified here is a builder rather than an operator? Should this read 

```suggestion
   The ``Any()`` builder uses the ``$elemMatch`` query operator.
```

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,331 @@
...
+Element Operators
+-----------------
+
+Element operators query data based on the presence or type of a field.
+
+For a full list of element operators, see the :manual:`Element
+Query Operators </reference/operator/query-element/>` page.
+
+The following example uses builders to create a query filter that matches all
+documents that have a ``rating`` field.
```
biniona-mongodb, line 270: Issue: [use colons at the end of sentence that introduces example](https://www.mongodb.com/docs/meta/style-guide/style/punctuation/#colons). Applies whenever applicable.

```suggestion
documents that have a ``rating`` field:
```

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,331 @@
...
+      { "_id" : 3, "make" : "PRS", "models" : ["Silver Sky", "SE", "Custom"], "establishedYear" : 1985, "rating" : 9 }
+      { "_id" : 5, "make" : "Ibanez", "models" : ["RG", "AZ"], "establishedYear" : 1957, "rating" : 7 }
+
+To learn more about using builders, see :ref:`csharp-builders`.
+
+Evaluation Operators
+--------------------
+
+Evaluation operators analyze data on individual fields, or on the entire collection's
+documents. Common evaluation operators include ``Regex()`` and ``Text()``.
```
biniona-mongodb, line 293: Issue: I believe the evaluation operators are $regex and $text rather than the builder methods for these operators.

Suggestion: Clarify language around operators and builders where applicable. A potential solution could be to make a table that lists the operator and the corresponding builder when it makes sense.

source/includes/fundamentals/code-examples/specify-query/FindAndBuilder.cs
```diff
@@ -0,0 +1,8 @@
+var builder = Builders<Guitar>.Filter;
```
biniona-mongodb, line 1: Suggestion: To help keep the examples and output in sync, you could have all examples in a single file and then use literal includes to include only the desired lines in each example. You could also generate the output blocks for code examples from the output of this one file.

biniona-mongodb (changes requested):
Looking good, a few more things to consider

source/fundamentals/crud/read-operations/specify-query.txt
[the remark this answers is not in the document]
jordan-smith721: I can make a ticket to update these to add these refs since they don't currently exist, and then update them on this page.

source/fundamentals/crud/read-operations/specify-query.txt
jordan-smith721: Looks like I missed this specific one somehow. Fixed

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,331 @@
...
+Overview
+--------
+
+In this guide, you can learn how to specify a query to match a subset
+of documents using the {+driver-long+}.
+
+Most CRUD (Create, Read, Update, Delete) operations allow you to narrow the set
+of matched documents by passing a **query filter** as an argument to the driver
+method. A query filter is an expression that specifies the documents you want to
+retrieve in your query.
```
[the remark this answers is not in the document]
jordan-smith721: I don't think query and query filter are interchangeable. Query is the entire statement (like `collection.Find(filter)`) and query filter is just the filter that's passed to the query.

I reformatted the paragraph to specify the operations as you mentioned in your suggestions. I also reworded it to hopefully make more sense and show the distinction between the terms "query" and "query filter"

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,331 @@
...
+
+Most CRUD (Create, Read, Update, Delete) operations allow you to narrow the set
+of matched documents by passing a **query filter** as an argument to the driver
+method. A query filter is an expression that specifies the documents you want to
+retrieve in your query. 
+
+.. note:: Using LINQ
+
+   You can also specify queries using LINQ. To learn more about using LINQ, see
+   :ref:`csharp-linq`.
```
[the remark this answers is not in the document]
jordan-smith721: LINQ is a whole other method of querying on it's own, so I don't think it falls under query filter as it's talked about here. I reworded the admonition to reflect that difference and give context on what the "also" refers to.

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,331 @@
...
+   * - ``!=``
+     - ``Ne()``
+     - Not equal to
+
+For a full list of comparison operators, see the :manual:`Comparison
+Query Operators </reference/operator/query-comparison/>` page.
+
+The following example specifies a query filter as a parameter to the ``Find()``
+method, using the ``Guitar`` class. The query matches all documents where the
+``establishedYear`` field is greater than ``1985``.
```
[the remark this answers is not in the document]
jordan-smith721: I removed "using the ``Guitar`` class from these paragraphs since we now state at the top that the ``Guitar``class is used in the examples for automapping. I left the rest of the paragraph the same as I think it's important to still have the focus of the example be specifying the query since that's the topic of the page, rather than the result of the query.

source/fundamentals/crud/read-operations/specify-query.txt
jordan-smith721: The queries end up the same, but the query filter and the way you construct the query are different, so they aren't identical. 
I changed the wording to show that the filters match the same documents in both, hopefully that helps accomplish what you are saying here.

source/fundamentals/crud/read-operations/specify-query.txt
jordan-smith721: Match documents is a better term here - changed!

I didn't take your exact suggested change here, but tried to edit the sentence to convey the same meaning.

source/fundamentals/crud/read-operations/specify-query.txt
jordan-smith721: Yes, changed to "match documents"

source/fundamentals/crud/read-operations/specify-query.txt
jordan-smith721: Changed the wording for the given examples

source/includes/fundamentals/code-examples/specify-query/FindAndBuilder.cs
jordan-smith721: Ideally I would like to have just one file for all examples (like we do in usage examples). Unfortunately io-code-block directives don't support the :start-after: and :end-before: options yet, so I had to separate the examples out in order to use that directive.

source/fundamentals/crud/read-operations/specify-query.txt
```diff
@@ -0,0 +1,353 @@
...
+The following ``Guitar`` class models the documents in this collection.
+
+.. TODO: Add link to mapping page.
+
+.. literalinclude:: /includes/fundamentals/code-examples/specify-query/Guitar.cs
+   :language: csharp
+   :copyable:
+   :dedent:
+
+The following code instantiates the ``guitars`` collection with this class. The
```
biniona-mongodb, line 51: Issue: The ``guitars`` collection is not created/instantiated through the following code snippet, but rather the ``_guitarsCollection`` object.

Suggestion: "with the class" is a bit unhelpful (it doesn't say what instantiating the class in this way). Consider using language like "serialize" / "deserialize":

```suggestion
The following code instantiates the ``Collection`` object ``_guitarsCollection`` using the ``Guitars`` class as a type parameter.  This type parameter  causes the ``_guitarsCollection`` object to automatically serialize and deserialize the BSON documents it receives from MongoDB to instances of the ``Guitars`` class: 
```

biniona-mongodb (approved):
LGTM

PR #51, opened 2022-12-05
Usage Examples

jordan-smith721:
### At the docs level:

- Are any topics missing?

### On each page:

- Is anything missing?


source/includes/code-examples/DeleteOne.cs
```diff
@@ -0,0 +1,77 @@
...
+public class DeleteOne
+{
+    private static IMongoCollection<Restaurant> _restaurantsCollection;
+    private static string _mongoConnectionString = "<Your MongoDB URI>";
+
+    public static void Main(string[] args)
+    {
+        Setup();
+
+        var doc = _restaurantsCollection.Find(Builders<Restaurant>.Filter
```
DmitryLukyanov, line 18: I doubt such formatting is correct, at the very least it should be:

     var doc = _restaurantsCollection.Find(
          Builders<Restaurant>.Filter.Eq(r => r.Name, "Ready Penny Inn")).First();

to not split logically the same line about definition filter and his configuration on 2 different line

DmitryLukyanov: but do we want to define more straight rules (not necessary the same as we use in the driver)? cc: @JamesKovacs

source/includes/code-examples/DeleteOne.cs
```diff
@@ -0,0 +1,77 @@
...
+        Restore(doc);
+    }
+
+    private static DeleteResult DeleteARestaurantBuilder()
+    {
+        // start-delete-one-builders
+        var filter = Builders<Restaurant>.Filter
+            .Eq(r => r.Name, "Ready Penny Inn");
+
+        var result = _restaurantsCollection.DeleteOne(filter);
```
DmitryLukyanov, line 36: better just to use `return _restaurantsCollection.DeleteOne(filter);` Same in other places

source/includes/code-examples/FindMany.cs
```diff
@@ -0,0 +1,97 @@
...
+            .Where(r => r.Cuisine == "Pizza").ToList();
+        // end-find-linq-sync
+
+        WriteLine("Number of documents found: " + query.Count);
+    }
+
+    private static void FindAllRestaurantsSync()
+    {
+        // start-find-all-sync
+        var restaurants = _restaurantsCollection.Find(new BsonDocument()).ToList();
```
DmitryLukyanov, line 61: for empty filter, I would recommend using: `EmptyFilterDefinition<BsonDocument>.Empty`. Same in other places

source/includes/code-examples/FindOne.cs
```diff
@@ -0,0 +1,85 @@
...
+    {
+        // start-find-builders
+        var filter = Builders<Restaurant>.Filter
+            .Eq(r => r.Name, "Bagels N Buns");
+
+        var restaurant = _restaurantsCollection.Find(filter).FirstOrDefault();
+        // end-find-builders
+
+        WriteLine(restaurant.ToBsonDocument());
+
```
DmitryLukyanov, line 40: blank line

source/includes/code-examples/FindManyAsync.cs
```diff
@@ -0,0 +1,97 @@
...
+        // end-find-builders-async
+    }
+
+    private static async Task<List<Restaurant>> FindMultipleRestaurantsLINQAsync()
+    {
+        // start-find-linq-async
+        return await _restaurantsCollection.AsQueryable()
+            .Where(r => r.Cuisine == "Pizza").ToListAsync();
+        // end-find-linq-async
+
```
DmitryLukyanov, line 56: blank line

source/includes/code-examples/FindOneAsync.cs
```diff
@@ -0,0 +1,84 @@
...
+
+    private static async Task<Restaurant> FindOneRestaurantBuilderAsync()
+    {
+        // start-find-builders
+        var filter = Builders<Restaurant>.Filter
+            .Eq(r => r.Name, "Bagels N Buns");
+
+        return await _restaurantsCollection.Find(filter).FirstOrDefaultAsync();
+        // end-find-builders
+
```
DmitryLukyanov, line 41: blank line

source/includes/code-examples/InsertOneAsync.cs
```diff
@@ -0,0 +1,89 @@
...
+        InsertOneRestaurant();
+
+        // find and print newly inserted document
+        var document = _restaurantsCollection.Find(Builders<Restaurant>.Filter
+            .Eq(r => r.Name, "Mongo's Pizza")).FirstOrDefault();
+
+        WriteLine("Document Inserted: " + document.ToBsonDocument());
+
+        Cleanup();
+
```
DmitryLukyanov, line 28: blank line

source/includes/code-examples/UpdateOneAsync.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        var filter = Builders<Restaurant>.Filter
+            .Eq(restaurant => restaurant.Name, oldValue);
+
+        var update = Builders<Restaurant>.Update
+            .Set(restaurant => restaurant.Name, newValue);
+
+        return await _restaurantsCollection.UpdateOneAsync(filter, update);
+        // end-update-one-async
+
```
DmitryLukyanov, line 42: blank line

source/includes/code-examples/delete-many/DeleteManyAsync.cs
```diff
@@ -0,0 +1,58 @@
...
+public class DeleteManyAsync
+{
+    private static IMongoCollection<Restaurant> _restaurantsCollection;
+    private static string _mongoConnectionString = "<Your MongoDB URI>";
+
+    public static void Main(string[] args)
+    {
+        Setup();
+
+        var docs = _restaurantsCollection.Find(Builders<Restaurant>.Filter
```
DmitryLukyanov, line 18: do we really need regex condition for simple deleteMany operation?

source/includes/code-examples/insert-many/InsertMany.cs
```diff
@@ -0,0 +1,95 @@
...
+        // Establish the connection to MongoDB and get the restaurants database
+        var mongoClient = new MongoClient(MongoConnectionString);
+        var restaurantsDatabase = mongoClient.GetDatabase("sample_restaurants");
+        _restaurantsCollection = restaurantsDatabase.GetCollection<Restaurant>("restaurants");
+    }
+
+    private static List<Restaurant> GenerateDocuments()
+    {
+        // Generate 5 new restaurant documents
+        List<Restaurant> restaurantsList = new List<Restaurant>();
```
DmitryLukyanov, line 67: var restaurantsList = new List<Restaurant>();

same in other places

source/includes/code-examples/update-many/UpdateMany.cs
```diff
@@ -0,0 +1,86 @@
...
+                .Eq(CuisineField, NewCuisine);
+
+            var update = Builders<Restaurant>.Update
+                .Set(restaurant => restaurant.Cuisine, OldCuisine);
+
+            _restaurantsCollection.UpdateMany(filter, update);
+        }
+    }
+}
+
```
DmitryLukyanov, line 85: blank lines

source/includes/code-examples/update-many/UpdateManyAsync.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            var update = Builders<Restaurant>.Update
+                .Set(restaurant => restaurant.Cuisine, OldCuisine);
+
+            _restaurantsCollection.UpdateMany(filter, update);
+        }
+    }
+}
+
+
```
DmitryLukyanov, line 86: blank lines

DmitryLukyanov (commented):
Do we want to mention BulkWrite methods like https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver/IMongoCollection.cs#L154?

source/includes/code-examples/DeleteOne.cs
JamesKovacs: For readability, I would suggest moving the builder expression into a `filter` variable the same way as we did it for `DeleteOne` below.

source/includes/code-examples/FindMany.cs
JamesKovacs: I would suggest `Builders<BsonDocument>.Filter.Empty` is more idiomatic and discoverable.

source/includes/code-examples/delete-many/DeleteManyAsync.cs
mongoKart: Would a text search ($text) achieve a similar result more efficiently? If not, I think I'll need to rework the whole example.

mongoKart:
> Do we want to mention BulkWrite methods like https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver/IMongoCollection.cs#L154?

Based on my notes from our initial meetings with James, we're going to add bulk operations later, not as part of the MVP.

source/includes/code-examples/FindOneAsync.cs
```diff
@@ -0,0 +1,82 @@
...
+        WriteLine(linqDocument);
+    }
+
+    private static async Task<Restaurant> FindOneRestaurantBuilderAsync()
+    {
+        // start-find-builders
+        var filter = Builders<Restaurant>.Filter
+            .Eq(r => r.Name, "Bagels N Buns");
+
+        return await _restaurantsCollection.Find(filter).FirstOrDefaultAsync();
```
BorisDog, line 39: Should we use `FindAsync` in Async example?

source/includes/code-examples/DeleteOneAsync.cs
```diff
@@ -0,0 +1,75 @@
...
+        Setup();
+
+        var doc = _restaurantsCollection.Find(Builders<Restaurant>.Filter
+            .Eq(r => r.Name, "Ready Penny Inn")).First();
+
+        // Delete a document using builders
+        WriteLine("Deleting a document with builders...");
+        var result = DeleteARestaurantBuilderAsync();
+
+        WriteLine($"Deleted documents: {result.Result.DeletedCount}");
```
BorisDog, line 25: Except for a few cases, it's a bad practice to wait for task with `Task.Result`.
I'd suggest to make Main async:   `async Task Main(string[] args)` and 
`var deleteRersult = await DeleteARestaurantBuilderAsync();` in all async examples.

source/includes/code-examples/FindOneAsync.cs
DmitryLukyanov: nope, I think I asked it somewhere before. Find (fluent) and FindSync/FindAsync are different approaches and as I recall at least initially we wanted to mention only fluent approach

source/includes/code-examples/DeleteOneAsync.cs
DmitryLukyanov: Agreed. Main should be async

source/includes/code-examples/FindMany.cs
```diff
@@ -0,0 +1,98 @@
...
+        var restaurants = _restaurantsCollection.Find(filter).ToList();
+        // end-find-builders-sync
+
+        WriteLine("Number of documents found: " + restaurants.Count);
+    }
+
+    private static void FindMultipleRestaurantsLINQSync()
+    {
+        // start-find-linq-sync
+        var query = _restaurantsCollection.AsQueryable()
```
DmitryLukyanov, line 51: the result of this line is not a query. It's already restaurants

source/includes/code-examples/FindManyAsync.cs
```diff
@@ -0,0 +1,97 @@
...
+using static System.Console;
+
+namespace CSharpExamples.UsageExamples;
+
+public class FindManyAsync
+{
+    private static IMongoCollection<Restaurant> _restaurantsCollection;
+    private static string _mongoConnectionString = "<Your MongoDB URI>";
+
+    public static void Main(string[] args)
```
DmitryLukyanov, line 15: Main should be async

source/includes/code-examples/FindMany.cs
DmitryLukyanov: Better if your sync and async code will be the same. Here you're adding printing, but that doesn't happen here:  https://github.com/mongodb/docs-csharp/pull/51/files#diff-21844f77d63bb9dceb89de54218ec0150222045850393b20081c6c6dee52f3e4R40

source/includes/code-examples/FindManyAsync.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        // Find all documents
+        WriteLine("Finding all documents...:");
+        var allRestaurants = FindAllRestaurantsAsync();
+        WriteLine("Number of documents found: " + allRestaurants.Result.Count);
+    }
+
+    private static async Task<List<Restaurant>> FindMultipleRestaurantsBuilderAsync()
+    {
+        // start-find-builders-async
```
DmitryLukyanov, line 42: What does `Builders` mean in this method name?

source/includes/code-examples/FindMany.cs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// start-model
+public class Restaurant
+{
+    public ObjectId Id { get; set; }
+
+    public string Name { get; set; }
+
+    [BsonElement("restaurant_id")]
```
DmitryLukyanov, line 87: I'm not sure why do we define a BsonElement for a single field. I think, it should be either for all or nothing?

source/includes/code-examples/FindManyAsync.cs
mongoKart: It's meant to differentiate it from the LINQ version of the method.

source/includes/code-examples/FindMany.cs
mongoKart: I don't know definitively. @BorisDog or @rstam, how should we use [BsonElement] in this class?

DmitryLukyanov: I know how we can use it :) , the question is why we define it only for one field? I would expect either defining it for all props or for no one

mongoKart: Oh, sorry Dima! 

When I remove the attribute from that property, I get a runtime error, but when I add it to just that one property, it works. You're saying it's still good practice to add it everywhere, even if it isn't strictly required?

DmitryLukyanov: this attribute just says that a db value in `restaurant_id` field will be mapped into `RestaurantId`. Exception you saw previously said that this mapping is not configured, but it's only because you already had `restaurant_id` in your collection. An empty collection won't give this issue

DmitryLukyanov: note: I don't say we need to remove it, I simple mean we should choose a single approach

source/includes/code-examples/DeleteOneAsync.cs
```diff
@@ -0,0 +1,75 @@
...
+public class DeleteOneAsync
+{
+    private static IMongoCollection<Restaurant> _restaurantsCollection;
+    private static string _mongoConnectionString = "<Your MongoDB URI>";
+
+    public static async Task Main(string[] args)
+    {
+        Setup();
+
+        var doc = _restaurantsCollection.Find(Builders<Restaurant>.Filter
```
DmitryLukyanov, line 18: same as in some other comment, let's move this filter into different variable

source/includes/code-examples/FindMany.cs
```diff
@@ -0,0 +1,96 @@
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Attributes;
+using MongoDB.Bson.Serialization.Conventions;
+using MongoDB.Driver;
+using static System.Console;
```
DmitryLukyanov, line 5: is it really needed? Sounds like you use different formats with `Console.WriteLine` in different places. I would recommend using the most standard: `Console.Writeline` (not just WriteLine + line 5)

source/includes/code-examples/FindMany.cs
```diff
@@ -0,0 +1,96 @@
...
+        // start-find-linq-async
+        return _restaurantsCollection.AsQueryable()
+            .Where(r => r.Cuisine == "Pizza").ToList();
+        // end-find-linq-async
+    }
+
+    private static void FindAllRestaurantsSync()
+    {
+        // start-find-all-sync
+        var restaurants = _restaurantsCollection.Find(Builders<BsonDocument>.Filter.Empty())
```
DmitryLukyanov, line 60: let's move filter into different variable

source/includes/code-examples/FindManyAsync.cs
```diff
@@ -0,0 +1,97 @@
...
+            .Eq(r => r.Cuisine, "Pizza");
+
+        return await _restaurantsCollection.Find(filter).ToListAsync();
+        // end-find-builders-async
+    }
+
+    private static async Task<List<Restaurant>> FindMultipleRestaurantsLINQAsync()
+    {
+        // start-find-linq-async
+        return await _restaurantsCollection.AsQueryable()
```
DmitryLukyanov, line 53: nit: in the driver we would use this formatting:

     return await _restaurantsCollection
            .AsQueryable()
            .Where(r => r.Cuisine == "Pizza")
            .ToListAsync(); 

where each operation is placed to an own raw, but feel free to use a different format if you think it suites better

source/includes/code-examples/FindManyAsync.cs
```diff
@@ -0,0 +1,97 @@
...
+            .Where(r => r.Cuisine == "Pizza").ToListAsync();
+        // end-find-linq-async
+    }
+
+    private static async Task<List<Restaurant>> FindAllRestaurantsAsync()
+    {
+        // start-find-all-async
+        return await _restaurantsCollection.Find(Builders<BsonDocument>.Filter.Empty())
```
DmitryLukyanov, line 61: let's move filter to a new variable

source/includes/code-examples/insert-many/InsertMany.cs
```diff
@@ -0,0 +1,95 @@
...
+            restaurantsList.Add(newRestaurant);
+        }
+
+        return restaurantsList;
+    }
+
+    private static void Cleanup()
+    {
+        var filter = Builders<Restaurant>.Filter.Eq(r => r.Name, "Mongo's Pizza");
+
```
DmitryLukyanov, line 92: nit: no need for blank line as I understand

source/includes/code-examples/InsertOne.cs
```diff
@@ -0,0 +1,88 @@
...
+        ConventionRegistry.Register("CamelCase", camelCaseConvention, type => true);
+
+        // Establish the connection to MongoDB and get the restaurants database
+        var mongoClient = new MongoClient(_mongoConnectionString);
+        var restaurantsDatabase = mongoClient.GetDatabase("sample_restaurants");
+        _restaurantsCollection = restaurantsDatabase.GetCollection<Restaurant>("restaurants");
+    }
+
+    private static void Cleanup()
+    {
```
DmitryLukyanov, line 66: I think it would be good to make sure that CleanUp has the same format everywhere (so move filter into a diff variable)

DmitryLukyanov: same in some other places in this PR

source/includes/code-examples/update-many/UpdateMany.cs
DmitryLukyanov: not fixed

source/includes/code-examples/update-many/UpdateManyAsync.cs
```diff
@@ -0,0 +1,86 @@
...
+    public class UpdateManyAsync
+    {
+        private static IMongoCollection<Restaurant> _restaurantsCollection;
+        private const string MongoConnectionString = "<Your MongoDB URI>";
+
+        private const string OldCuisine = "Pizza";
+        private const string NewCuisine = "Pasta and breadsticks";
+        private const string CuisineField = "cuisine";
+
+        public static void Main(string[] args)
```
DmitryLukyanov, line 16: should be Async

source/includes/code-examples/update-many/UpdateManyAsync.cs
```diff
@@ -0,0 +1,86 @@
...
+            // Number of restaurants with new cuisine
+            WriteLine($"Restaurants with {CuisineField} \"{NewCuisine}\" found after update: {FindCountOfRestaurantsWithCuisine(NewCuisine)}");
+
+            // Reset sample data
+            Write("Resetting sample data...");
+            ResetSampleData();
+            WriteLine("done.");
+        }
+
+        private static async Task<UpdateResult> UpdateManyRestaurants()
```
DmitryLukyanov, line 39: should be `UpdateManyRestaurantsAsync`

source/includes/code-examples/update-many/UpdateManyAsync.cs
DmitryLukyanov: not fixed

source/includes/code-examples/delete-many/DeleteMany.cs
```diff
@@ -0,0 +1,57 @@
...
+public class DeleteMany
+{
+    private static IMongoCollection<Restaurant> _restaurantsCollection;
+    private static string _mongoConnectionString = "<Your MongoDB URI>";
+
+    public static void Main(string[] args)
+    {
+        Setup();
+
+        var docs = _restaurantsCollection.Find(Builders<Restaurant>.Filter
```
DmitryLukyanov, line 18: No need for regex? I guess it's just comparing full value?

DmitryLukyanov: Move filter into diff variable

source/includes/code-examples/update-many/UpdateMany.cs
```diff
@@ -0,0 +1,86 @@
+using MongoDB.Bson.Serialization.Conventions;
+using MongoDB.Driver;
+using static System.Console;
+
+namespace UsageExamples.UpdateMany
+{
+    public class UpdateMany
+    {
+        private static IMongoCollection<Restaurant> _restaurantsCollection;
+        private const string MongoConnectionString = "<Your MongoDB URI>";
```
DmitryLukyanov, line 10: use the same naming: `_variableName` and variable definition (static vs const) approaches. Same in other places

source/includes/code-examples/DeleteOneAsync.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        return _restaurantsCollection.DeleteOneAsync(filter);
+        // end-delete-one-builders-async
+    }
+
+    private static void Restore(Restaurant doc)
+    {
+        _restaurantsCollection.InsertOne(doc);
+    }
+    private static void Setup()
```
DmitryLukyanov, line 44: add a blank line between methods

source/includes/code-examples/insert-many/InsertMany.cs
```diff
@@ -0,0 +1,95 @@
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Attributes;
+using MongoDB.Bson.Serialization.Conventions;
+using MongoDB.Driver;
+using static System.Console;
+
+namespace CsharpExamples.UsageExamples.InsertMany;
```
DmitryLukyanov, line 7: let's use the same namespace naming approach everywhere

source/includes/code-examples/FindManyAsync.cs
mongoKart: I like this idea, but because it would need to be made across all LINQ pages, I'd like to ticket this separately. I'll make the ticket and we can implement the change after these docs are live.

source/includes/code-examples/insert-many/InsertMany.cs
mongoKart: Do you mean the blank line between defining the filter and calling DeleteMany()? That's just for readability.

source/includes/code-examples/update-many/UpdateMany.cs
mongoKart: Could you clarify this? I understand that we should be consistent across all code files, but do you also mean that the code you highlighted here needs to be changed?

source/includes/code-examples/insert-many/InsertMany.cs
DmitryLukyanov: If "2 things" are logically about the same step, I would say they don't need any separator and vice-versa if 2 lines are about different steps to do something, then it makes sense to visually split them. But feel free to choose the way you prefer, I think it's not important

source/includes/code-examples/update-many/UpdateMany.cs
DmitryLukyanov: you use 2 diff approaches to define connection string in this PR. The above one and the way when you use `static readonly string __connectionString`. I think we need only one.

source/includes/code-examples/FindMany.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        Console.WriteLine();
+
+        // Find all restaurants
+        Console.WriteLine("Finding all documents...:");
+        restaurants = FindAllRestaurantsSync();
+        Console.WriteLine("Number of documents found: " + restaurants.Count);
+    }
+
+    public static List<Restaurant> FindMultipleRestaurantsBuilderSync()
```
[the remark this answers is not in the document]
mongoKart: "Builders" and "LINQ" were the terms we used when talking to James during the planning phase, so that's why we used them here. We only used "Builder" to differentiate these methods from those that use LINQ. 

I brought this up in the #docs-csharp channel to see if anyone has any other thoughts. This would probably also be a follow-up ticket, since it affects multiple docs pages.

source/includes/code-examples/insert-many/InsertMany.cs
mongoKart: I have standardized all namespace declarations to use the above format.

source/includes/code-examples/replace-one/Restaurant.cs
```diff
@@ -0,0 +1,24 @@
+using MongoDB.Bson;
+using MongoDB.Bson.Serialization.Attributes;
+
+namespace CSharpExamples.UsageExamples.ReplaceOne;
+
+//start-model
+public class Restaurant
```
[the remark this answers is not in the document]
mongoKart: We had three different writers working on these examples and ended up doing them differently. AFAIK, best practice is to be put classes in their own files. I'll make that change for all examples.

source/includes/code-examples/FindMany.cs
mongoKart: This isn't a functional issue and affects multiple pages in the docs, so I'll create a separate ticket to be addressed after the docs are released.

source/includes/code-examples/insert-many/InsertMany.cs
mongoKart: We've added a blank line everywhere else, so I'll leave this for now.

source/includes/code-examples/FindMany.cs
DmitryLukyanov: The key point as I mentioned before is that you use 2 different formats (which way better is a different question). See: https://github.com/mongodb/docs-csharp/pull/51/files#diff-790e8dbf2076f80693f7b76b1e5c2603abce672af63c9d44f000e6ccb1ba1fc7R24 and https://github.com/mongodb/docs-csharp/pull/51/files#diff-57fa48a1757ed5fb36ac3003227f186eae73dd0fba0ef4b36c5c56999e8ae7f5R19. In this particular file, you don't need this configuration at all.

source/includes/code-examples/find-many/FindManyAsync.cs
```diff
@@ -0,0 +1,77 @@
...
+using MongoDB.Driver;
+using MongoDB.Driver.Linq;
+using static System.Console;
+
+namespace CSharpExamples.UsageExamples.FindMany;
+
+public class FindManyAsync
+{
+    private static IMongoCollection<Restaurant> _restaurantsCollection;
+    private static string _mongoConnectionString = "<Your MongoDB URI>";
```
DmitryLukyanov, line 13: this is not changed

source/includes/code-examples/find-one/FindOneAsync.cs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    private static void Setup()
+    {
+        // This allows automapping of the camelCase database fields to our models. 
+        var camelCaseConvention = new ConventionPack { new CamelCaseElementNameConvention() };
+        ConventionRegistry.Register("CamelCase", camelCaseConvention, type => true);
+
+        // Establish the connection to MongoDB and get the restaurants database
+        var mongoClient = new MongoClient(_mongoConnectionString);
```
DmitryLukyanov, line 58: `_mongoConnectionString` is not defined in this file anymore

source/includes/code-examples/update-many/UpdateManyAsync.cs
```diff
@@ -0,0 +1,85 @@
...
+        Setup();
+
+        // Extra space for console readability 
+        WriteLine();
+
+        // Number of restaurants with old cuisine
+        WriteLine($"Restaurants with {CuisineField} \"{OldCuisine}\" found: {FindCountOfRestaurantsWithCuisine(OldCuisine)}");
+
+        // Update many documents synchronously
+        var asyncResult = UpdateManyRestaurantsAsync();
```
DmitryLukyanov, line 27: should be `await`

source/includes/code-examples/FindMany.cs
mongoKart: Updated all `WriteLine` calls to `Console.WriteLine` and removed `using static System.Console`

source/includes/code-examples/find-many/FindManyAsync.cs
mongoKart: D'oh! VS Code search functionality isn't serving me well today.

DmitryLukyanov (commented):
Sounds like some of the provided examples are not compiled. Can you build all these examples to ensure that all works as expected?

source/includes/code-examples/find-one/FindOneAsync.cs
```diff
@@ -0,0 +1,61 @@
...
+{
+    private static IMongoCollection<Restaurant> _restaurantsCollection;
+    private const string MongoConnectionString = "<Your MongoDB URI>";
+
+    public static async Task Main(string[] args)
+    {
+        Setup();
+
+        // Find one document using builders
+        var buildersDocument = await FindOneRestaurantBuilderAsync().ToBsonDocument();
```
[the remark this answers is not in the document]
mongoKart: Yes, I had the same idea. I put all code examples in Rider and ensured they would all compile.

BorisDog (approved):
LGTM

source/includes/code-examples/insert-many/InsertMany.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            Restaurant newRestaurant = new()
+            {
+                Name = "Mongo's Pizza",
+                RestaurantId = $"12345-{i}",
+                Cuisine = "Pizza",
+                Address = new BsonDocument
+                {
+                    {"street", "Pizza St"},
+                    {"zipcode", "10003"},
```
DmitryLukyanov, line 77: nit: remove `,` after zipcode

source/usage-examples/insertOne.txt
```diff
@@ -0,0 +1,89 @@
...
+Expected Result
+~~~~~~~~~~~~~~~
+
+After running either of the preceding full examples, the ``InsertOne()``
+method inserts the document, and the :ref:`Find() <csharp-find-one>` method returns
+the newly inserted document. The output is as follows:
+
+.. code-block:: none
+
+   // results truncated
```
DmitryLukyanov, line 73: I read this:
image
like the commented text was added in output which is not true, can we formulate it better?

source/includes/code-examples/FindMany.cs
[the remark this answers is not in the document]
DmitryLukyanov: this is not addressed. I still think that the provided naming is misleading. I would prefer to resolve this, but will rely on you

DmitryLukyanov (approved):
See some remaining comments and then LGTM

mongoKart:
@DmitryLukyanov I'm happy to keep discussing the naming conventions with the team and seeing if we can agree on an alternative. Because we use the same Builders/LINQ distinction in other docs, I don't mind using it here for the time being.

PR #52, opened 2022-12-05
Connection Pages

jordan-smith721:
### At the docs level:

- Are any topics missing?

### On each page:

- Is anything missing?


source/fundamentals/connection/connection-options.txt
```diff
@@ -0,0 +1,418 @@
...
+       | **Default**: ``false``
+
+   * - **ApplicationName**
+     - ``appName=yourApp``
+     - | The app name the driver passes to the server in the client metadata as part of 
+         the connection handshake. The server prints this value to the MongoDB logs once 
+         the connection is established. The value is also recorded in the slow query logs 
+         and profile collections.
+       |
+       | **Data Type**: {+string-data-type+}
```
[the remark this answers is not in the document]
mongoKart: That's not a convention across docs. I'll change it here.

source/fundamentals/connection/connection-options.txt
```diff
@@ -0,0 +1,418 @@
...
+         that the server supports. See :ref:`csharp-network-compression` for 
+         more information. 
+       | 
+       | **Data Type**: ``CompressorConfiguration``
+       | **Default**: ``null``
+
+   * - **ConnectTimeout**
+     - ``connectTimeoutMS=0``
+     - | The length of time the driver tries to establish a single TCP socket connection 
+         to the server before timing out. Specifying ``0`` disables the connection timeout.
```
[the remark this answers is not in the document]
mongoKart: Do you mean setting `connectTimeout=TimeSpan.Zero` will actually set the timeout to 30 seconds in the driver? Or do you mean that the writing incorrectly suggests that it will?

source/fundamentals/connection/connection-options.txt
```diff
@@ -0,0 +1,418 @@
...
+     - ``readConcernLevel=local``
+     - | The client's default read concern. 
+         See :ref:`read concern <read-concern>` for more information.
+       |
+       | **Data Type**: ``ReadConcern``
+       | **Default**: ``ReadConcern.Default``
+
+   * - **ReadEncoding**
+     - {+not-available+}
+     - | Specifies whether UTF-8 string deserialization is strict or lenient.
```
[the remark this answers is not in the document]
mongoKart: Could you provide a little more information about what you'd like to see here? It sounds like there are multiple possible UTF-8 encodings that could be used here, and Strict Encoding (the default), will throw an exception on invalid bytes. Is that right?

source/fundamentals/connection/connection-options.txt
```diff
@@ -0,0 +1,418 @@
...
+     - ``mongodb+srv://``
+     - | Specifies whether to use the standard connection string format (``MongoDB``) 
+         or the DNS seed list format (``MongoDBPlusSrv``). 
+         See :manual:`the MongoDB Manual</reference/connection-string/>` for more 
+         information about connection string formats. 
+       |
+       | **Data Type**: ``ConnectionStringScheme``
+       | **Default**: ``ConnectionStringScheme.MongoDB``
+      
+   * - **SdamLogFilename**
```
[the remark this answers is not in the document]
mongoKart: If it's obsolete in version 2.19, we should remove it.

source/fundamentals/connection/connection-options.txt
```diff
@@ -0,0 +1,418 @@
...
+         a scheme of ``\"mongodb+srv\"`` or specify other TLS options, 
+         this option defaults to ``true``.
+       |
+       | **Data Type**: {+bool-data-type+}
+       | **Default**: ``false``
+
+   * - **WaitQueueTimeout**
+     - ``waitQueueTimeoutMS=0``
+     - | The length of time the driver tries to check out a connection from a 
+         server's connection pool before timing out. Specifying ``0`` means no timeout.
```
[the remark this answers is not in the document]
mongoKart: Do we want users to use `TimeSpan.Infinite` instead of `0`? If so, I'll remove this sentence from the docs. I only included it because I thought it was unintuitive and wanted people to be aware.

source/fundamentals/connection/connection-options.txt
```diff
@@ -0,0 +1,418 @@
...
+         and profile collections.
+       |
+       | **Data Type**: {+string-data-type+}
+       | **Default**: ``null``
+
+   * - **AutoEncryptionOptions**
+     - {+not-available+}
+     - | Settings for automatic client-side encryption.
+       |
+       | **Data Type**: ``AutoEncryptionOptions``
```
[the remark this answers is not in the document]
mongoKart: Good idea.

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,27 @@
...
+
+==========
+Connection
+==========
+
+.. default-domain:: mongodb
+
+.. toctree::
+
+   /fundamentals/connection/connect
```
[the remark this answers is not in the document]
mongoKart: That's not a bad idea, but if we did, we'd probably put it on an entirely separate page. It's not something we decided to include in the MVP, so we can make it a separate ticket for release after Jan. 10.

source/fundamentals/connection/tls.txt
```diff
@@ -0,0 +1,228 @@
...
+
+You can configure your X.509 certificate using ``MongoClientSettings``. The following
+code sample creates a new X.509 certificate object using the certificate file named 
+``client.p12``, which is protected by the password ``mySuperSecretPassword``. The code
+then adds this certificate to the ``SslSettings.ClientCertificates`` array in 
+``MongoClientSettings``.
+
+.. code-block:: csharp
+
+   var cert = new X509Certificate2("client.p12", "mySuperSecretPassword");
```
mongoKart, line 67: In [another PR](https://github.com/mongodb/docs-csharp/pull/57#discussion_r1045291627), Dima suggested using .p12 certs instead of .pfx here. Can you confirm that my updates are correct?

source/fundamentals/connection/connection-options.txt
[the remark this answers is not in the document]
BorisDog: Strict and lenient encoding are just shortcuts defined in `MongoDB.Bson.IO.Utf8Encodings`.
Not sure how it's best to indicate the default value, as `Strict` word would not be informative for a user.
So we could just say that default encoding will throw on invalid bytes, or/and link to our custom `Utf8Encodings.Strict`.
Maybe @rstam will have a better idea.

source/fundamentals/connection/tls.txt
BorisDog: @DmitryLukyanov please :)

DmitryLukyanov: as wiki [says](https://en.wikipedia.org/wiki/PKCS_12) `.p12` is the successor to `pfx` format. Nowadays these formats are interchangeable. As far as I know microsoft for historical reasons still prefers pfx extension, but I would say better to mention the most modern format :) We also use `p12` format in our tests

source/fundamentals/connection/connect.txt
```diff
@@ -0,0 +1,123 @@
...
+MongoDB instance on port ``27017`` of ``localhost``:
+
+.. literalinclude:: /includes/fundamentals/code-examples/connection/MongoClientSettings.cs
+   :language: csharp
+   :dedent:
+
+Other Connection Targets
+------------------------
+
+Connect to a Standalone Deployment
```
DmitryLukyanov, line 87: If by `standalone deployment` you mean a single mongod instance, then it's not really true. The same connection string will work for sharded cluster too

source/fundamentals/connection/connect.txt
```diff
@@ -0,0 +1,123 @@
...
+.. literalinclude:: /includes/fundamentals/code-examples/connection/ReplicaSetConnection.cs
+   :language: csharp
+
+
+
+
+
+
+
+
```
DmitryLukyanov, line 123: blank lines

source/fundamentals/connection/connect.txt
```diff
@@ -0,0 +1,123 @@
...
+.. tip::
+
+   If your deployment is hosted on Atlas, follow the
+   :atlas:`Atlas driver connection guide </driver-connection?tck=docs_driver_nodejs>`
+   to retrieve your connection string.
+
+Connect to a Replica Set
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+In the following example, the driver uses a sample connection URI to connect to the MongoDB replica set ``sampleRS``, which is running on port ``27017`` of ``sample.host1``, ``sample.host2``, and ``sample.host3``:
```
DmitryLukyanov, line 112: this is a bit confusing. It can be read like if you want to use rs connection string, you should provide all servers in the connection string which is not true. I think the main using mode is providing a single any rs host and `replicaSet` name

source/fundamentals/connection/connection-options.txt
```diff
@@ -0,0 +1,414 @@
...
+
+   * - **HeartbeatTimeout**
+     - {+not-available+}
+     - | The length of time a monitoring socket can be idle before timing out.
+       |
+       | **Data Type**: ``TimeSpan``
+       | **Default**: Same value as ``ConnectTimeout``
+     
+   * - **IPv6**
+     - {+not-available+}
```
DmitryLukyanov, line 171: ipv6 can be set via connectionString [option](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs#L1011)

source/fundamentals/connection/connection-options.txt
```diff
@@ -0,0 +1,414 @@
...
+     - | Specifies whether the host address is in IPv6 format.
+       | 
+       | **Data Type**: {+bool-data-type+}
+       | **Default**: ``false``
+
+   * - **IsFrozen**
```
DmitryLukyanov, line 177: Not sure whether we need this, but even if so, I would mention that this is readonly field

source/fundamentals/connection/connection-options.txt
```diff
@@ -0,0 +1,414 @@
...
+  If you're using a query parameter for a time duration, the value must be in 
+  milliseconds. For example, to specify 60 seconds, use the value ``60000``. If you're 
+  using a ``MongoClientSettings`` object for a time duration, use the appropriate 
+  ``TimeSpan`` value.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 40 30 30
+
+   * - **MongoClientSettings** Property
```
DmitryLukyanov, line 70: no logging configuration for mongoClientSettings?

source/fundamentals/connection/connection-options.txt
```diff
@@ -0,0 +1,414 @@
...
+   * - **HeartbeatInterval**
+     - ``heartbeatFrequencyMS=5000``
+     - | The interval between regular server-monitoring checks. Must be greater than or 
+         equal to 500 milliseconds.
+       |
+       | **Data Type**: ``TimeSpan``
+       | **Default**: 10 seconds
+
+   * - **HeartbeatTimeout**
+     - {+not-available+}
```
DmitryLukyanov, line 164: it's possible to configure it via connection [string](https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs#L1008)

source/fundamentals/connection/connection-options.txt
```diff
@@ -0,0 +1,414 @@
...
+       | **Default**: ``WriteConcern.Acknowledged``
+
+   * - **WriteEncoding**
+     - {+not-available+}
+     - | Specifies whether UTF-8 string serialization is strict or lenient. With strict
+         encoding, the driver will throw an exception when it encounters an invalid 
+         UTF-8 byte sequence.
+       |
+       | **Data Type**: ``UTF8Encoding``
+       | **Default**: Strict encoding
```
DmitryLukyanov, line 414: missed https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs#L1163

DmitryLukyanov: https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs#L1139

DmitryLukyanov: https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs#L1139

DmitryLukyanov: it's only few I noticed after quick look. Not all of options we have in connectionString needs to be mentioned here, but maybe it makes sense to review it again?

source/fundamentals/connection/connection-options.txt
```diff
@@ -0,0 +1,414 @@
...
+     - ``tlsDisableCertificateRevocationCheck=false``
+     - | TLS/SSL options, including client certificates, revocation handling, and 
+         enabled and disabled TLS/SSL protocols.
+       |
+       | **Data Type**: `SslSettings <{+api-root+}/T_MongoDB_Driver_SslSettings.htm>`__
+       | **Default**: ``null``
+
+   * - **UseTls**
+     - ``tls=true`` or ``ssl=true``
+     - | Specifies whether to require TLS for connections to the server. If you use
```
DmitryLukyanov, line 377: formatting:
image

source/fundamentals/connection/connection-options.txt
```diff
@@ -0,0 +1,414 @@
...
+.. list-table::
+   :header-rows: 1
+   :widths: 40 30 30
+
+   * - **MongoClientSettings** Property
+     - Connection URI Example
+     - Description
+
+   * - **AllowInsecureTls**
+     - ``tlsInsecure=true``
```
DmitryLukyanov, line 75: should we mention restrictions on some options combinations like here https://github.com/mongodb/mongo-csharp-driver/blob/master/src/MongoDB.Driver.Core/Core/Configuration/ConnectionString.cs#L1139 (and in nearest lines)?

source/fundamentals/connection/connection-options.txt
BorisDog: Good catch.

source/fundamentals/connection/connect.txt
mongoKart: Removed. RST syntax requires a single blank line at the bottom of the file, which I left in place.

source/fundamentals/connection/connection-options.txt
mongoKart: Can you explain?

source/fundamentals/connection/connect.txt
DmitryLukyanov: I'm ok to remove it

source/fundamentals/connection/connection-options.txt
DmitryLukyanov: Currently MongoClientSettings provides a way to configure logging. We probably need to describe it

DmitryLukyanov:
Btw, it can be a good idea to add a section about a load balancing mode. I guess it's a good idea for a different ticket

source/fundamentals/connection/connection-options.txt
mongoKart: When I wrote this page, I included only the options that are part of the Connection String spec--hence, a few were left out. Below are the connection-string options in the driver code that I left out the first time.

- `connect` - [The corresponding MongoClientSettings property appears to be obsolete.](https://mongodb.github.io/mongo-csharp-driver/2.18/apidocs/html/P_MongoDB_Driver_Core_Configuration_ConnectionString_Connect.htm)

- `fsync` - Added. The MongoClientSettings property is read-only, but it looks like the connection-string option can set this to true/false. Is that right?

- `gssapiservicename` - How does this interact with the `authMechanismProperties` option? That option is covered as part of the `Credential` property.

- `maxlifetimems` - Added.

- `safe` - What does this option do? It appears to relate to the write concern, but I don't see how. And what's the corresponding `MongoClientSettings` property?

- `uuidrepresentation` - Added, though some things suggest `GuidRepresentation` is obsolete; is that true? Can you confirm the default is `GuidRepresentation.CSharpLegacy`?

- `waitqueuemultiple` - [Obsolete.](https://mongodb.github.io/mongo-csharp-driver/2.18/apidocs/html/P_MongoDB_Driver_Core_Configuration_ConnectionString_WaitQueueMultiple.htm)

- `waitqueuesize` - [Obsolete.](https://mongodb.github.io/mongo-csharp-driver/2.18/apidocs/html/P_MongoDB_Driver_Core_Configuration_ConnectionString_WaitQueueSize.htm)

- `zlibcompressionlevel` - I believe @JamesKovacs suggested not including zlib as a compression option. This is also why the `Compressors` property includes only snappy and zstd.

source/fundamentals/connection/connection-options.txt
mongoKart: Can you point me to some information?

source/fundamentals/connection/connection-options.txt
mongoKart: I've added some. Let me know if you think it's enough.

mongoKart:
> Btw, it can be a good idea to add a section about a load balancing mode. I guess it's a good idea for a different ticket

I created a new ticket for this.

source/fundamentals/connection/connection-options.txt
DmitryLukyanov: cc: @BorisDog

source/fundamentals/connection/connection-options.txt
DmitryLukyanov: * `connect ` - not needed
* `fsync` - it's part of writeConcern. Connection string allows specifying writeconcern options separately, but in MongoClientSettings we set full object (with all options including fsync) at once, so it's not readonly. To assign it you should do something close to: `clientSettings.WriteConcern = clientSettings.WriteConcern.With(fsync: false)`. I would say we need this since otherwise we will not have a way to set this option via connectionSring, but I'm not sure why it's not part of the uri spec..
* `gssapiservicename` - sounds like it's legacy option that is not part of the spec. The provided value actually just add a new authMechanism properties = `{ "SERVICE_NAME", value }`. I think we can ignore it.
* `safe` - I believe you can ignore it
* `uuidrepresentation` - this is only option for GuidRepresentationMode.V2. I believe we can ignore it since we're going to remove v2 mode in 3.0 driver release
* waitqueuemultiple / waitqueuesize  - I'm ok with not including it 
* `zlibCompressionLevel` - is part of the uri spec, so we really don't need this?

@JamesKovacs do you have any comments?

source/fundamentals/connection/connection-options.txt
BorisDog: Yes, `LoggingSettings` property.
https://mongodb.github.io/mongo-csharp-driver/2.18/reference/driver/logging/

source/fundamentals/connection/connection-options.txt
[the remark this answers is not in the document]
BorisDog: I think it's fine now.

source/fundamentals/connection/connection-options.txt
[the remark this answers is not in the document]
BorisDog: Looks good.

BorisDog (changes requested):
Looks good overall, we need to add MongoClientSettings.LoggingSettings option.

source/fundamentals/connection/connection-options.txt
JamesKovacs: Agreed on all those points. None of the noted URI options needs to be included in our docs. Most of those listed are deprecated. Even the ones not deprecated should only be used in very limited (and questionable) circumstances.

Regarding `gssapiServiceName`, see https://jira.mongodb.org/browse/DOCS-9897 for an explanation of its deprecation. Basically what Dima already said but in the form of a DOCS ticket.

DmitryLukyanov: what about `zlibCompressionLevel`? I remember some discussions about zlib compression, but not sure whether it was decided to deprecate this option(compression?) at some point?

DmitryLukyanov: note: see recent [ticket](https://jira.mongodb.org/browse/DOCSP-26706) for mongosh with a similar goal

source/fundamentals/connection/connection-options.txt
mongoKart: Leaving to a [later ticket](https://jira.mongodb.org/browse/DOCSP-26773), as mentioned by Jordan [here.](https://mongodb.slack.com/archives/C03RFE0V9NV/p1673306249775959?thread_ts=1673305947.079499&cid=C03RFE0V9NV)

source/fundamentals/connection/connection-options.txt
mongoKart: Leaving `zlib` out of docs, per [this discussion.](https://mongodb.slack.com/archives/C03RFE0V9NV/p1673477778917129)

source/fundamentals/connection/connection-options.txt
BorisDog: I would still suggest to add the `LoggingSettings` property even with most basic description like "Client logging settings". Just have it visible for the users, as it's quite valuable monitoring tool.

Agreed that we can have the more detailed examples and info done in later ticket.

BorisDog (changes requested):

BorisDog (approved):
LGTM!

PR #58, opened 2022-12-05
Class Mapping

jordan-smith721:
### At the docs level:

- Are any topics missing?

### On each page:

- Is anything missing?

### DIFF
--- a/source/fundamentals/class-mapping.txt
+++ b/source/fundamentals/class-mapping.txt
@@ -309,7 +309,7 @@ You can also specify the method while registering a class map:
    {
        classmap.AutoMap();
        classmap.MapMember(p => c.Age).SetShouldSerializeMethod(
-           obj => ((Person) obj).Age > 0
+           obj => ((Person) obj).Age != 0
        );
    });
 
--- a/source/fundamentals/class-mapping.txt
+++ b/source/fundamentals/class-mapping.txt
@@ -31,7 +31,7 @@ the name of the field in the document to the name of the property in the class.
    The type of the property in your class should match the type of the field in
    the document. The {+driver-short+} instantiates a serializer based on the
    type of the property in your class. If the types don't match when the driver
-   attempts to serialize the data, the serializer will throw an exception.
+   serializes the data, the serializer throws an exception.
 
 Manually Creating A Class Map
 -----------------------------
@@ -165,7 +165,7 @@ The following example shows how a document from the ``Person`` class with the
 
 .. code-block:: json
 
-   { "_id": "...", "_t": "personClass", "name": "...", "age": "...", "hobbies": [...]}
+   { "_id": "...", "_t": "personClass", "Name": "...", "Age": "...", "Hobbies": [...]}
 
 .. TODO: Link to page on polymorphism/discriminators
 
@@ -277,14 +277,15 @@ Dynamically Serialize Properties
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 You can use a method to determine whether or not to serialize a property. For
-the driver to automatically use the method when serializing, the method must be
-named ``ShouldSerialize<property name>``. When the driver sees a method with
-this naming convention, it will use the method to determine whether or not to
-serialize properties of the provided property name.
+the driver to automatically use the method when serializing, you must prefix the
+method name with ``ShouldSerialize`` followed by the name of the property that
+the method applies to. When the driver sees a method with this naming
+convention, it uses that method to determine whether or not to serialize
+properties that have the provided property name.
 
 The following example creates a method that only serializes the ``Age`` property
-if its value is greater than ``0``. The driver ignores any properties whose
-values don't meet this requirement:
+if its value is not equal to ``0``. The driver does not serialize any properties
+whose values don't meet this requirement:
 
 .. code-block:: csharp
 
@@ -296,7 +297,7 @@ values don't meet this requirement:
 
        public bool ShouldSerializeAge()
        {
-          return Age > 0;
+          return Age != 0;
        }
    }
 
--- a/source/fundamentals/class-mapping.txt
+++ b/source/fundamentals/class-mapping.txt
@@ -55,11 +55,11 @@ class:
 
 .. code-block:: csharp
    
-   BsonClassMap.RegisterClassMap<Person>(classmap => 
+   BsonClassMap.RegisterClassMap<Person>(classMap => 
    {
-       classmap.MapMember(p => p.Name);
-       classmap.MapMember(p => p.Age);
-       classmap.MapMember(p => p.Hobbies);
+       classMap.MapMember(p => p.Name);
+       classMap.MapMember(p => p.Age);
+       classMap.MapMember(p => p.Hobbies);
    });
 
 .. important::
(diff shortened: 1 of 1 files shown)


source/fundamentals/class-mapping.txt
```diff
@@ -0,0 +1,328 @@
...
+MongoDB collection, the {+driver-short+} automatically creates a **class map**
+that it uses to serialize or deserialize your data. It does this mapping by matching 
+the name of the field in the document to the name of the property in the class.
+
+.. important::
+
+   The type of the property in your class should match the type of the field in
+   the document. The {+driver-short+} instantiates a serializer based on the
+   type of the property in your class. If the types don't match when the driver
+   attempts to serialize the data, the serializer will throw an exception.
```
rstam, line 34: "attempts to deserialize the data"

It will always succeed when serializing.

source/fundamentals/class-mapping.txt
```diff
@@ -0,0 +1,328 @@
...
+   });
+
+In BSON, discriminators have the field name ``_t``. 
+
+The following example shows how a document from the ``Person`` class with the
+"personClass" discriminator appears in the collection after serialization:
+
+.. code-block:: json
+
+   { "_id": "...", "_t": "personClass", "name": "...", "age": "...", "hobbies": [...]}
```
rstam, line 168: The element names will match the property names: `Name`, `Age` and `Hobbies`.

source/fundamentals/class-mapping.txt
```diff
@@ -0,0 +1,328 @@
...
+.. TODO: Link to page on polymorphism/discriminators
+
+Mapping with Constructors
+-------------------------
+
+By default, the {+driver-short+} can automatically map a class only if the class has
+a constructor with no arguments. If you want the driver to use a constructor that accepts
+one or more arguments, you can add the ``BsonConstructor`` attribute to the constructor.
+In this case, the driver parses the expression tree to determine how to map the   
+constructor arguments to class properties or fields.
```
rstam, line 179: "parses the expression tree" is not an accurate statement

"examines the type using reflection" is more accurate

source/fundamentals/class-mapping.txt
```diff
@@ -0,0 +1,328 @@
...
+       {
+           Name = name;
+           Age = age;
+       }
+   }
+
+   BsonClassMap.RegisterClassMap<Person>(classmap =>
+   {
+       classmap.AutoMap();
+       classmap.MapCreator(p => new Person(p.Name, p.Age));
```
rstam, line 229: I don't think `MapCreator` used like this will work.

Have you verified it?

Normally you would pass `MapCreator` a static factory method.

source/fundamentals/class-mapping.txt
```diff
@@ -0,0 +1,328 @@
...
+   After the driver serializes a class with extra elements back to BSON, the 
+   extra elements may not be in the same order as they were in the original 
+   document.
+
+Dynamically Serialize Properties
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can use a method to determine whether or not to serialize a property. For
+the driver to automatically use the method when serializing, the method must be
+named ``ShouldSerialize<property name>``. When the driver sees a method with
```
rstam, line 281: Angle brackets used like this could be confusing in C# because that looks like a generic method.

Not sure if there is a better alternative though.

source/fundamentals/class-mapping.txt
```diff
@@ -0,0 +1,328 @@
...
+this naming convention, it will use the method to determine whether or not to
+serialize properties of the provided property name.
+
+The following example creates a method that only serializes the ``Age`` property
+if its value is greater than ``0``. The driver ignores any properties whose
```
rstam, line 286: "ignores any Age property values"

This only applies to serialization, not deserialization.

Also, normally this would be used to omit the default value, not a whole range of values (like negative ages). So I would expect to see `Age != 0` rather than `Age > 0`.

Of course if all ages are positive it amounts to the same thing. But I'd still rather see `!=` rather than `>`.

source/fundamentals/class-mapping.txt
rstam: There is also a `MapConstructor` method but it requires passing it a `ConstructorInfo` value which you would have to get using reflection.

source/fundamentals/class-mapping.txt
jordan-smith721: Good to know. Would it be sufficient to say "the driver examines the types to determine how to map the   
constructor arguments to class properties or fields", to avoid confusing anyone not familiar with reflection? 

I could also mention reflection and link to more information about it if we think it's necessary to mention.

source/fundamentals/class-mapping.txt
jordan-smith721: I took this code block from the current documentation here:
https://mongodb.github.io/mongo-csharp-driver/2.17/reference/bson/mapping/#creation-customization
(just changing the property names to match the other examples on this page). Is this information incorrect, or did I perhaps mistranslate it?

source/fundamentals/class-mapping.txt
jordan-smith721: Changed to `Age != 0` and specified that the driver only ignores during serialization

source/fundamentals/class-mapping.txt
rstam: Examines the types is a good way of putting it.

source/fundamentals/class-mapping.txt
rstam: I'll have to write a test to find out. I'll get back to you.

source/fundamentals/class-mapping.txt
rstam: Are you going to make this change?

source/fundamentals/class-mapping.txt
```diff
@@ -0,0 +1,329 @@
...
+   You must register a class map *before* it's needed in your code. We recommend
+   registering class maps prior to initializing a connection with MongoDB.
+
+You can also manually map a subset of class properties, while still
+allowing the driver to automatically map the remaining properties. To do this, 
+register a class map and call the ``AutoMap()`` method before manually 
+specifying your properties.
+
+In the following code example, the ``AutoMap()`` method maps all properties
+of the ``Person`` class except ``Hobbies``, which is mapped manually:
```
rstam, line 76: This is not quite right. Automap still maps all properties (including Hobbies).

What you can do is alter the mapping after Automap runs.

This example is a no-op because `classMap.MapMember(p => p.Hobbies)` does nothing because `Hobbies` is already mapped.

rstam: For this example to be meaningful the line after `AutoMap` should do something do change the mapping that `AutoMap` produced.

For example:

    classMap.MapMember(p => p.Hobbies).SetElementName("hobbies");

which of course is also a no-op if you assume a non-default set of conventions with camel case registered (which still concerns me), so maybe:

    classMap.MapMember(p => p.Hobbies).SetElementName("favorite_hobbies");

source/fundamentals/class-mapping.txt
```diff
@@ -0,0 +1,329 @@
...
+       public List<string> Hobbies {get; set;}
+
+       public Person(string name, string age)
+       {
+           Name = name;
+           Age = age;
+       }
+   }
+
+   BsonClassMap.RegisterClassMap<Person>(classmap =>
```
rstam, line 226: C# coding conventions use camel case: `classMap` (capital M)

rstam (changes requested):

source/fundamentals/class-mapping.txt
jordan-smith721: Oops, thought I did that earlier. It's changed now!

source/fundamentals/class-mapping.txt
jordan-smith721: Fixed

source/fundamentals/class-mapping.txt
jordan-smith721: Thanks, that does seem like a better example.

I've updated the wording for the last sentence before the example Let me know if it is more correct that way

rstam (approved):
LGTM

PR #74, opened 2023-02-03
DOCSP-27770 - Remove Placeholders

mongoKart:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-27770
Staging - https://docs-mongodborg-staging.corp.mongodb.com/csharp/docsworker-xlarge/docsp-27770-placeholder-bug/

## Self-Review Checklist


source/quick-start.txt
```diff
@@ -14,6 +14,49 @@ Quick Start
 
 .. include:: /includes/quick-start/overview.rst
 
+Create a MongoDB Cluster
```
mongoKart, line 17: Moved this section first so it goes from Create Cluster > Create Project > Code Project, rather than Create Project > Create Cluster > Code Project.

source/quick-start.txt
```diff
@@ -14,6 +14,49 @@ Quick Start
...
+------------------------
+
+Set Up a Free Tier Cluster in Atlas
```
mongoKart, line 20: Moved a lot of this content directly into this page instead of an `includes` file, since it's not used anywhere else.

source/quick-start.txt
```diff
@@ -14,6 +14,49 @@ Quick Start
...
+To retrieve the connection string for your Atlas cluster and user, complete the 
+:guides:`Get Connection String </atlas/connection-string>`
+guide. Once you have finished the guide, you will have a connection string similar to the 
+following:
+
+.. code-block:: bash
+
+   "mongodb+srv://username:password@cluster0.abc.mongodb.net/?retryWrites=true&w=majority"
+
+Save your Atlas connection string to an environment variable named ``MONGODB_URI``.
```
mongoKart, line 52: Moved this step to immediately after they retrieve the string, rather than having them copy it down somewhere else first.

source/quick-start.txt
```diff
@@ -14,6 +14,49 @@ Quick Start
...
+other connection options.
+
+To retrieve the connection string for your Atlas cluster and user, complete the 
+:guides:`Get Connection String </atlas/connection-string>`
+guide. Once you have finished the guide, you will have a connection string similar to the 
+following:
+
+.. code-block:: bash
+
+   "mongodb+srv://username:password@cluster0.abc.mongodb.net/?retryWrites=true&w=majority"
```
jordan-smith721, line 50: Suggestion: Use brackets here to make the placeholders stand out more 

```suggestion
   "mongodb+srv://<username>:<password>@cluster0.abc.mongodb.net/?retryWrites=true&w=majority"
```

source/quick-start.txt
jordan-smith721: I think might be confusing here for some users since they haven't created the app or the project directory yet. If I were starting off from scratch and reading this I wouldn't know if this should be in the same location as the app or somewhere else entirely.
 
I also wonder if all users will know what exactly we mean by saying to "save to an environment variable", but that might be a separate discussion.

source/quick-start.txt
```diff
@@ -39,32 +82,21 @@ Use the ``dotnet add`` command to add the {+driver-short+} to your project as a
...
+In this step, you'll use the {+driver-short+} 
+to connect to your MongoDB cluster and run a query on the sample data.
 
-.. code-block:: bash
-
-   export MONGODB_URI='<your atlas connection string>'
-
-.. important::
-
-   Make sure to replace the "<username>" and "<password>" sections of the connection string with the username and password of your Atlas user.
```
jordan-smith721, line 58: Suggestion: Could we add a version of this admonition to where we tell the user to create their environment variable? I think it's still good to point this out in case some users are just copy/pasting and don't realize the placeholders are there.

jordan-smith721 (changes requested):
I like the change to link out to the guides on how to use Atlas, rather than duplicating that information on our own pages. Just have a couple comments/suggestions on the new structure.

source/quick-start.txt
mongoKart: Yeah, the "environment variable" language was already here, but we can talk about whether to clarify it. 

I don't think it makes a difference where they run this command, since it's an environment-wide variable. What if we explain the steps a little more?

jordan-smith721 (approved):
LGTM

source/quick-start.txt
```diff
@@ -14,6 +14,52 @@ Quick Start
...
+To retrieve the connection string for your Atlas cluster and user, complete the 
+:guides:`Get Connection String </atlas/connection-string>`
+guide. Once you have finished the guide, you will have a connection string similar to the 
+following:
+
+.. code-block:: bash
+
+   "mongodb+srv://<username>:<password>@cluster0.abc.mongodb.net/?retryWrites=true&w=majority"
+
+Run the following code at the command prompt to save your Atlas connection string to an
```
ccho-mongodb, line 51: Suggestion:

What do you think about creating a new section for this part, perhaps entitled "Set Your Connection String Environment Variable"? That helps logically separate it from "Find Your Connection String".

If a user reads https://www.mongodb.com/docs/guides/atlas/account/ (suggested link in an earlier comment) to completion, they should be able to start with their "connection string in [their] copy buffer".

source/quick-start.txt
```diff
@@ -39,32 +85,21 @@ Use the ``dotnet add`` command to add the {+driver-short+} to your project as a
...
+Open the file named ``Program.cs`` in the base directory of your project. Copy the
 following sample code into ``Program.cs`` 
 
 .. literalinclude:: /includes/quick-start/Program.cs
    :language: csharp
    :dedent:
 
-The sample code runs a query against your sample dataset in MongoDB Atlas. Run the sample code with the following command from your command line:
+The sample code runs a query against your sample dataset in MongoDB Atlas. Run the 
+sample code with the following command from your command line:
```
ccho-mongodb, line 102: Suggestion:

1) I prefer using "this" over "the" when referring to an instance of something introduced just prior because it's more specific.
2) Sometimes there's a fine line between possessive "your" and non-possessive "the".
3) Reorder the final sentence so "command" is closest to the command itself which can be helpful for understanding its purpose.

```suggestion
This sample code runs a query against your sample dataset in MongoDB Atlas. Run it
from your command line using the following command:
```

source/quick-start.txt
```diff
@@ -81,7 +116,12 @@ After completing this step, you should have a working application that uses
 the {+driver-short+} to connect to your MongoDB cluster, run a query on the
 sample data, and print out the result.
 
+To learn more about connecting to Atlas with the {+driver-short+}, see
+the :atlas:`Atlas driver connection </driver-connection>` guide
+and select **{+language+}** from the *Select your language* dropdown.
```
ccho-mongodb, line 121:
```suggestion
and select **{+language+}** from the :guilabel`Select your language` dropdown.
```

source/quick-start.txt
```diff
@@ -14,6 +14,52 @@ Quick Start
...
 
+Create a MongoDB Cluster
+------------------------
+
+Set Up a Free Tier Cluster in Atlas
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Complete the
+:atlas:`Get Started with Atlas </getting-started>` guide to set up a new
+Atlas account and free tier MongoDB cluster, load datasets, and
```
ccho-mongodb, line 25:
```suggestion
Atlas account and free tier MongoDB cluster, load sample datasets, and
```

source/quick-start.txt
```diff
@@ -14,6 +14,52 @@ Quick Start
...
 .. include:: /includes/quick-start/overview.rst
 
+Create a MongoDB Cluster
+------------------------
+
+Set Up a Free Tier Cluster in Atlas
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Complete the
+:atlas:`Get Started with Atlas </getting-started>` guide to set up a new
```
ccho-mongodb, line 24: Issue:

This linked guide actually currently shows everything that this guide shows, including running CRUD operations. Therefore, if you complete that guide, there's no reason to continue with this one.

Suggestion:
You might consider linking to this one instead for now since it ends with "Get Connection String": https://www.mongodb.com/docs/guides/atlas/account/

ccho-mongodb (changes requested):
The reorder-ing makes a lot of sense, but I think there are some issues stemming from the linked Atlas guide, or at least how the content should flow if they follow that guide to completion.

source/quick-start.txt
ccho-mongodb: I think the language should probably also use the guilabel per the style guide. I'm not sure how useful using a source constant is here since it should match the Atlas UI representation which could change.

source/quick-start.txt
mongoKart: Changed "using" to "by using" as required by Style Guide, but adopted all other suggested changes

source/quick-start.txt
```diff
@@ -14,6 +14,41 @@ Quick Start
...
 .. include:: /includes/quick-start/overview.rst
 
+Create a MongoDB Cluster
+------------------------
+
+Set Up a Free Tier Cluster in Atlas
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Complete Chapter 1 of the
+:guides:`MongoDB Guide </atlas/account>` to set up a new
```
mongoKart, line 24: I didn't know whether there's a standard way to refer to this guide (guides?). In this case, we want the user to complete just Chapter 1.

source/quick-start.txt
```diff
@@ -81,7 +105,12 @@ After completing this step, you should have a working application that uses
 the {+driver-short+} to connect to your MongoDB cluster, run a query on the
 sample data, and print out the result.
 
+To learn more about connecting to Atlas with the {+driver-short+}, see
+the :atlas:`Atlas driver connection </driver-connection>` guide
+and select **{+language+}** from the :guilabel`Select your language` dropdown.
```
ccho-mongodb, line 110: Suggestion:

I think consistent usage of guilabel could help as suggested in a [previous comment](https://github.com/mongodb/docs-csharp/pull/74#discussion_r1097964746):

```suggestion
and select :guilabel:`C# / .NET` from the :guilabel`Select your language` dropdown.
```

ccho-mongodb (approved):
LGTM, added a few suggestions. Personally, I think the Chapter organization of the guides are not intuitive and isn't explained anywhere to my knowledge, so I would prefer to communicate the section of the guide in a way that doesn't require figuring it out.

source/quick-start.txt
mongoKart: Thought I committed that change last time. Must have gotten lost when reconciling commits on GitHub and VSCode.

PR #79, opened 2023-02-17
DOCSP-27919: stableapi connection for atlas

rustagir:
# Pull Request Info

[PR Reviewing Guidelines](https://github.com/mongodb/docs-golang/blob/master/REVIEWING.md)

JIRA - https://jira.mongodb.org/browse/DOCSP-27919
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/csharp/docsworker-xlarge/DOCSP-27919-connection-stableapi/fundamentals/connection/connect/#connect-to-atlas

## Self-Review Checklist

- [x] Is this free of any warnings or errors in the RST? (one error unrelated to this PR)
- [x] Did you run a spell-check?
- [ ] Did you run a grammar-check?
- [ ] Are all the links working?

PR #80, opened 2023-02-22
CC feedback on DOCSP-27919

rustagir:
https://docs-mongodbcom-staging.corp.mongodb.com/csharp/docsworker-xlarge/CC-feedback-27919/fundamentals/connection/connect/#connect-to-atlas

#79

### DIFF
--- a/source/fundamentals/connection/connect.txt
+++ b/source/fundamentals/connection/connect.txt
@@ -87,11 +87,16 @@ Other Connection Targets
 Connect to Atlas
 ~~~~~~~~~~~~~~~~
 
-In the following example, the driver uses a sample connection URI to
-connect to a MongoDB instance on Atlas, using placeholders for your
-username and password. In this example, the client uses
-the :ref:`Stable API <csharp-stable-api>` feature, though
-you can connect to Atlas without this feature.
+To connect to Atlas, create a client by passing your connection string
+and any settings. You can set a Stable API server version as a setting
+to avoid breaking changes when you upgrade to a new server version. To
+learn more about the Stable API feature, see the :ref:`Stable API page
+<csharp-stable-api>`.
+
+The following code shows how you can instantiate a client that uses a
+Stable API version and a sample connection string with placeholders for
+your username and password. Instantiating the client creates a
+connection to MongoDB:
 
 .. literalinclude:: /includes/fundamentals/code-examples/connection/AtlasConnection.cs
    :language: csharp


source/fundamentals/connection/connect.txt
```diff
@@ -87,11 +87,25 @@ Other Connection Targets
...
-connect to a MongoDB instance on Atlas, using placeholders for your
-username and password. In this example, the client uses
-the :ref:`Stable API <csharp-stable-api>` feature, though
-you can connect to Atlas without this feature.
+To connect to an Atlas deployment, you need to create a client. You can
+create a client that uses your connection string and other 
+client options by passing a ``MongoClientSettings`` object to the ``MongoClient``
+constructor. To specify your connection URI, pass it to the ``FromConnectionString()``
+method, creating a ``MongoClientSettings`` instance. To specify any other
+settings, set the relevant fields of the settings object.
```
ccho-mongodb, line 95: Suggestion:
I think it would be easier to read if the terminology was consistent (client options vs settings). I think in this case, it's more confusing to make the switch in terminology because the class name is MongoClientSettings.

```suggestion
client options, set the relevant fields of the MongoClientSettings object.
```

source/fundamentals/connection/connect.txt
```diff
@@ -87,11 +87,25 @@ Other Connection Targets
...
+
+To learn more about connection options, see the
+:ref:`Connection Options page <csharp-connection-options>`.
+
+You can set the {+stable-api+} version as a setting to avoid
```
ccho-mongodb, line 100: Suggestion:

I think using the verb "set" to describe an action on a "setting" can be confusing as there are other equivalent words. I think the current wording also attributes the "set" action to "as a setting" which is an incorrect relation and should be "the setting" instead.

Per another comment in this review, it could be helpful to stay consistent in terminology.

```suggestion
You can set the {+stable-api+} version client option to avoid
```

source/fundamentals/connection/connect.txt
```diff
@@ -87,11 +87,25 @@ Other Connection Targets
...
+breaking changes when you upgrade to a new server version. To 
+learn more about the {+stable-api+} feature, see the :ref:`{+stable-api+} page
+<csharp-stable-api>`.
+
+The following code shows how you can instantiate a client that uses the
+{+stable-api+} version and a sample connection string with placeholders for
+your username and password. Instantiating the client creates a
+connection to MongoDB:
```
ccho-mongodb, line 108: Issue:

1) I think the style guide mentions that when using a colon, it should introduce the content that follows it (style/punctuation). I think "Instantiating the client creates a connection to MongoDB:" is describing what happens when you instantiate a client rather than what the code example shows. I think the first sentence describes the code sample, but the second one separates it and therefore the colon is not suitable.

2) I think the conjunction is ambiguous in this sentence. One version reads as  "...code shows how you can... instantiate a client... and a sample connection string" and another as "... instantiate a client that uses the Stable API... and [uses] a sample connection string." I think the sentence structure can be changed to avoid the ambiguity.

Suggestion:

Perhaps simplifying the message could resolve both issues.

E.g.

```suggestion
The following code shows how you can specify the connection string and the Stable API client option when connecting to a MongoDB deployment:
```

ccho-mongodb (changes requested):
Thanks for incorporating prior feedback on the corresponding Golang page in your changes.

I think it could be helpful to clarify whether the drivers should present consistent code examples with Atlas UI team and update them if needed.

Feel free to reach out if you would like me to communicate with the Atlas UI team or after finding that information, if you want to pair on the solution.

source/includes/fundamentals/code-examples/connection/AtlasConnection.cs
```diff
@@ -1,10 +1,12 @@
 using MongoDB.Driver;
 
-// Connection URI
+// Replace the placeholders with your credentials
```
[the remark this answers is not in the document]
rustagir: Discussed with Atlas UI team. Successful confirmations do not need to be uniform across drivers, but are requested. I am adding a database command execution to the code to ping the `admin` database, which I believe all deployments on Atlas are created with.

On the other point-- they prefer individual placeholders for user/pass as opposed to a placeholder for the entire connection string. Will have to update in Go :/ but leaving as is in this PR.

source/fundamentals/connection/connect.txt
```diff
@@ -87,11 +87,25 @@ Other Connection Targets
...
-username and password. In this example, the client uses
-the :ref:`Stable API <csharp-stable-api>` feature, though
-you can connect to Atlas without this feature.
+To connect to an Atlas deployment, you need to create a client. You can
+create a client that uses your connection string and other 
+client options by passing a ``MongoClientSettings`` object to the ``MongoClient``
+constructor. To specify your connection URI, pass it to the ``FromConnectionString()``
+method, creating a ``MongoClientSettings`` instance. To specify any other
+settings, set the relevant fields of the settings object.
+
```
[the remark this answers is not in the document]
rustagir: Yes, it was getting a little dense with all the monospace object references

source/fundamentals/connection/connect.txt
rustagir: changed all references to options to settings

source/fundamentals/connection/connect.txt
rustagir: The code also includes a confirmation of successful connection now, so I will integrate this suggestion with that new description as well.

source/fundamentals/connection/connect.txt
```diff
@@ -87,19 +87,30 @@ Other Connection Targets
 Connect to Atlas
 ~~~~~~~~~~~~~~~~
 
-In the following example, the driver uses a sample connection URI to
-connect to a MongoDB instance on Atlas, using placeholders for your
-username and password. In this example, the client uses
-the :ref:`Stable API <csharp-stable-api>` feature, though
-you can connect to Atlas without this feature.
+To connect to a MongoDB deployment on Atlas, you need to create a client. You can
```
ccho-mongodb, line 90: Suggestion:

No changes necessary necessary, but technically, I think you can connect without creating a client. You can use some other tool or application in which a client is already created for you. Instead, this sentence could suggest the path of action using the following language:

```suggestion
To connect to a MongoDB deployment on Atlas, create a client. You can
```

source/includes/fundamentals/code-examples/connection/AtlasConnection.cs
```diff
@@ -1,11 +1,18 @@
...
 const string connectionUri = "mongodb+srv://<username>:<password>@cluster0.sample.mongodb.net/?retryWrites=true&w=majority";
 
-var serverApi = new ServerApi(ServerApiVersion.V1);
 var settings = MongoClientSettings.FromConnectionString(connectionString);
+
+// Set the ServerApi field of the settings object to Stable API version 1
+var serverApi = new ServerApi(ServerApiVersion.V1);
```
ccho-mongodb, line 10: Suggestion:
I think you probably don't need to set this to a variable. I'm not able to think of any advantages in splitting it into two lines since it's fairly short and the variable isn't used anywhere else.

I'd recommend combining lines 10 and 11 unless there's a good reason not to. E.g.:

```
settings.ServerApi = new ServerApi(ServerApiVersion.V1);
```

source/includes/fundamentals/code-examples/connection/AtlasConnection.cs
```diff
@@ -1,11 +1,18 @@
...
 settings.ServerApi = serverApi;
 
 // Create a new client and connect to the server
-var client = new MongoClient(settings);
\ No newline at end of file
+var client = new MongoClient(settings);
+
+// Send a ping to confirm a successful connection
+var result = client.GetDatabase("admin").RunCommand<BsonDocument>(new BsonDocument("ping", 1));
+Console.WriteLine(result.ToJson());
```
ccho-mongodb, line 18: Question:
I thought that the "successful confirmation" was going to be uniform as requested as mentioned in [this thread](https://github.com/mongodb/docs-csharp/pull/80#discussion_r1115970166). Or am I interpreting that response incorrectly?

Suggestion:
I think it would make sense to print a human-readable message that it was successful similar to [the one you added to the Go driver docs](https://github.com/mongodb/docs-golang/blob/master/source/includes/fundamentals/code-snippets/srv.go#L38) regardless of what operation is used to determine the connection.

ccho-mongodb (approved):
LGTM with a couple suggestions and question (potential issue).

PR #90, opened 2023-03-30
DOCSP-28590 Add POCOs to fully runnable examples

jordan-smith721:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-28590
Staging - Changes don't show in staging

## Self-Review Checklist

- [] Did you run a spell-check?

### DIFF
--- a/source/includes/code-examples/delete-many/DeleteMany.cs
+++ b/source/includes/code-examples/delete-many/DeleteMany.cs
@@ -54,3 +54,46 @@ public class DeleteMany
         _restaurantsCollection = restaurantsDatabase.GetCollection<Restaurant>("restaurants");
     }
 }
+
+public class Restaurant
+{
+    public ObjectId Id { get; set; }
+
+    public string Name { get; set; }
+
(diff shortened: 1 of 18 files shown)


DBirtolo-mdb (approved):
LGTM but with one question. Do you think it's worth moving these addition classes above the individual sample code? For example, moving it above the class `UpdateOneAsync`? That way readers know what those classes are when they hit them in the sample code?

jordan-smith721:
> LGTM but with one question. Do you think it's worth moving these addition classes above the individual sample code? For example, moving it above the class `UpdateOneAsync`? That way readers know what those classes are when they hit them in the sample code?

I think in this case we want the focus of the sample code to be on the methods rather than the classes. Adding the classes into these files is mostly just to ensure the full code compiles when it's copy/pasted. The user also will have had to scroll past the documentation explaining those classes on the usage example page before they can click on the link that holds this code, so I think it would be safe to assume they'll be aware of what they are.

PR #107, opened 2023-07-10
DOCSP-30836-URI-error

nickldp:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-30836>
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/csharp/docsworker-xlarge/DOCSP-30836-URI-error-message/)

## Self-Review Checklist

### DIFF
--- a/source/includes/quick-start/Program.cs
+++ b/source/includes/quick-start/Program.cs
@@ -4,7 +4,7 @@ using MongoDB.Bson;
 var connectionString = Environment.GetEnvironmentVariable("MONGODB_URI");
 if (connectionString == null)
 {
-    Console.WriteLine("You must set your 'MONGODB_URI' environmental variable. See\n\t https://www.mongodb.com/docs/drivers/csharp/current/quick-start/#set-your-connection-string");
+    Console.WriteLine("You must set your 'MONGODB_URI' environment variable. To learn how to set it, see https://www.mongodb.com/docs/drivers/csharp/current/quick-start/#set-your-connection-string");
     Environment.Exit(0);
 }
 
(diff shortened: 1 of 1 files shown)


source/includes/quick-start/Program.cs
```diff
@@ -4,7 +4,7 @@
 var connectionString = Environment.GetEnvironmentVariable("MONGODB_URI");
 if (connectionString == null)
 {
-    Console.WriteLine("You must set your 'MONGODB_URI' environmental variable. See\n\t https://www.mongodb.com/docs/drivers/go/current/usage-examples/#environment-variable");
+    Console.WriteLine("You must set your 'MONGODB_URI' environmental variable. See\n\t https://www.mongodb.com/docs/drivers/csharp/current/quick-start/#set-your-connection-string");
```
ccho-mongodb, line 7: Issue:
I think "environmental variable" is a typo and the correct term is "environment variable". See https://en.wikipedia.org/wiki/Environment_variable

Suggestion:

Use "environment variable" instead of "environmental variable".
I think it could also be good to use our writing conventions in error messages, in which case "see <link>" should provide some description of what to expect when they click/open it.

```suggestion
    Console.WriteLine("You must set your 'MONGODB_URI' environment variable. To learn how to set it, see https://www.mongodb.com/docs/drivers/csharp/current/quick-start/#set-your-connection-string");
```

ccho-mongodb (approved):
Thanks for the update! Left an issue & suggestion, but you should be ok to merge after addressing it.

PR #117, opened 2023-09-06
DOCSP-32483 programmatic tagging script for csharp repo

sarahemlin:
JIRA - <https://jira.mongodb.org/browse/DOCSP-32483>

Hi @mongoKart - here's the face.toml file for the csharp repo.  Please merge if everything looks good!

Thank you!
Sarah

### DIFF
--- a/source/facets.toml
+++ b/source/facets.toml
@@ -1,0 +1,7 @@
+[[facets]]
+category = "programming_language"
+value = "csharp"
+
+[[facets]]
+category = "target_product"
+value = "drivers"


source/facets.toml
```diff
@@ -0,0 +1,7 @@
+[[facets]]
+category = "programming_language"
+value = "csharp"
```
mongoKart, line 3: A minor point here: The C# DBX team has explicitly asked that we call this driver the .NET/C# driver, since it can technically work with most (all?) .NET languages. It's fine to call it csharp behind the scenes (you can see we did that with this repo name), but for display and search purposes, including ".NET" would be good.

I'm not sure if that should change here or elsewhere, so I'll wait on approving for now.

sarahemlin: Hey @mongoKart - this is very good to know, but a little bit different as we've split the tags by general product (drivers) and language (c#), but not driver name.  The facet values would be filters for search refinement only; you won't be able to filter by driver name but by a combination of driver & programming language which is important because we have more languages than drivers.  We don't have .NET as a programming language (approved by Rachelle & the leads), though.  We could add a manual keyword directive for .NET, but that would be outside the script (which doesn't apply because keywords aren't facet values).

mongoKart: Appreciate the context. Just wanted to be sure the driver is displayed appropriately.

mongoKart (approved):
lgtm

PR #120, opened 2023-09-25
DOCSP-26326 serialization

jordan-smith721:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-26326
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/csharp/docsworker-xlarge/DOCSP-26326-serialization/fundamentals/data-formats/serialization/

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/data-formats/serialization.txt
+++ b/source/fundamentals/data-formats/serialization.txt
@@ -13,65 +13,10 @@ Serialization
 Overview
 --------
 
-In this guide, you can learn about serialization with the {+driver-long+}.
-Serialization is the process of mapping a C# object to and from a BSON document.
-
-Serializer Registry
--------------------
-
-The serializer registry is an implementation of ``IBsonSerializerRegistry``
-that contains all registered serializers. You can access the serializer registry
-with the ``SerializerRegistry`` property of the ``BsonSerializer``
-class as follows:
-
-.. code-block:: csharp
-
-   var intSerializer = BsonSerializer.SerializerRegistry.GetSerializer<int>();
-
-.. important::
-
-   The serializer registry is a global registry. This means that you cannot use
-   multiple registries in a single application.
-
-Serialization Provider
-----------------------
-
-A serialization provider allows you to provide serializers to your
-serialization registry. The serialization provider should be registered as early
-as possible in your application to ensure that the serializers are used by the
-registry. Your serialization provider must implement the
-``IBsonSerializationProvider`` interface. 
-
-Example
-~~~~~~~
-
-The following example implements a serialization provider. The provider has a
-``GetSerializer`` method that returns a serializer based on the type of the
-input. If the method input is an ``int`` it returns a custom ``Int32``
-serializer, otherwise it returns ``null`` so that it can be handled by another
-provider. 
-
-.. code-block:: csharp
-   
-   class MyProvider : IBsonSerializationProvider
-   {
-       public IBsonSerializer GetSerializer(Type type)
-       {
-           if (type == typeof(int))
-           {
-               return new MyInt32Serializer();
-           }        
-   
-           return null;
-       }
-   }
-
-You can register the serialization provider using the
-``RegisterSerializationProvider`` method as shown in the following example:
-
-.. code-block:: csharp
-
-   BsonSerializer.RegisterSerializationProvider(new MyProvider);
+In this guide, you can learn how to use the {+driver-long+} to perform
+:wikipedia:`serialization </serialization>`.
+Serialization is the process of mapping a C# object to a BSON document for
+storage in MongoDB.
 
 Serializers
 -----------
@@ -84,35 +29,54 @@ primitive types, collection types, and custom classes.
 For a full list of available serializers, see the
 `Serializers namespace API documentation <{+api-root+}/N_MongoDB_Bson_Serialization_Serializers.htm>`__.
 
-Custom Serializers
-~~~~~~~~~~~~~~~~~~
+Serializer Registry
+-------------------
 
-In some cases, you may need to implement a custom serializer. When creating a
-custom serializer, your class should inherit from ``SerializerBase<T>`` and
-override the ``Deserialize()`` and ``Serialize()`` methods. 
-
-The following code example shows how to implement a custom serializer for
-an ``Int32``:
+The serializer registry is an implementation of the ``IBsonSerializerRegistry`` interface
+that contains all registered serializers. Before you can use a serializer, you
+must register it by adding it to the serializer registry, as
+shown in the following example:
 
 .. code-block:: csharp
 
-   class MyInt32Serializer : SerializerBase<int>
+   BsonSerializer.RegisterSerializer(typeof(MySerializerType), new MySerializer); 
+
+You can access the serializer registry with the ``SerializerRegistry`` property
+of the ``BsonSerializer`` class as follows:
+
+.. code-block:: csharp
+
+   var intSerializer = BsonSerializer.SerializerRegistry.GetSerializer<int>();
+
+.. important::
+
+   The serializer registry is a global registry. This means that you cannot use
+   multiple registries in a single application.
+
+Custom Serializers
+~~~~~~~~~~~~~~~~~~
+
+In some cases, you might need to create a custom serializer. When creating a
+custom serializer, implement the ``SerializerBase<T>`` interface and
+override the ``Deserialize()`` and ``Serialize()`` methods. 
+
+The following code example shows how to implement a custom serializer for
+an ``int``:
+
+.. code-block:: csharp
+
+   class MyIntSerializer : SerializerBase<int>
    {
        public override int Deserialize(BsonDeserializationContext context, BsonDeserializationArgs args)
        {
-           return context.Reader.ReadInt32();
+           return context.Reader.ReadInt();
        }
    
        public override void Serialize(BsonSerializationContext context, BsonSerializationArgs args, int value)
        {
-           context.Writer.WriteInt32(value);
+           context.Writer.WriteInt(value);
        }
    }
-
-.. important:: 
-
-   The preceding code example shows an overly simplistic serializer. When
-   creating a custom serializer ensure it handles all standard and edge cases.
 
 .. tip::
 
@@ -122,25 +86,25 @@ an ``Int32``:
 Opt-in Interfaces
 -----------------
 
-The {+driver-short+} has several opt-in interfaces that you can implement with
-your custom serializer based on type of data your serializer handles.
+The {+driver-short+} has several optional interfaces that your custom serializer
+class can implement, depending on the type of data the serializer handles.
 
 IBsonIdProvider
 ~~~~~~~~~~~~~~~
 
 The `IBsonIdProvider
 <{+api-root+}/T_MongoDB_Bson_Serialization_IBsonIdProvider.htm>`__
-interface provides the ``GetDocumentId`` and ``SetDocumentId``
-methods, and is useful if your class uses an ID type other than ``ObjectId``.
+interface provides the ``GetDocumentId()`` and ``SetDocumentId()``
+methods, and is useful if the object you are serializing uses an ID type other than ``ObjectId``.
 
 IBsonDocumentSerializer
 ~~~~~~~~~~~~~~~~~~~~~~~
 
 Implementing the `IBsonDocumentSerializer
 <{+api-root+}/T_MongoDB_Bson_Serialization_IBsonDocumentSerializer.htm>`__
-interface enables the driver to access your class's member
-information. This allows the driver to properly construct type-safe queries when using a
-custom serializer.
+interface enables the driver to access the member
+information of the object your are serializing. This allows the driver to
+properly construct type-safe queries when using a custom serializer.
 
 IBsonArraySerializer
 ~~~~~~~~~~~~~~~~~~~~
@@ -159,8 +123,5 @@ guide, see the following API documentation:
 - `IBsonSerializerRegistry <{+api-root+}/P_MongoDB_Bson_Serialization_BsonSerializer_SerializerRegistry.htm>`__
 - `SerializerRegistry <{+api-root+}/P_MongoDB_Bson_Serialization_BsonSerializer_SerializerRegistry.htm>`__
 - `BsonSerializer <{+api-root+}/T_MongoDB_Bson_Serialization_BsonSerializer.htm>`__
-- `IBsonSerializationProvider <{+api-root+}/T_MongoDB_Bson_Serialization_IBsonSerializationProvider.htm>`__
-- `GetSerializer <{+api-root+}/M_MongoDB_Bson_Serialization_IBsonSerializationProvider_GetSerializer.htm>`__
-- `RegisterSerializationProvider <{+api-root+}/M_MongoDB_Bson_Serialization_BsonSerializer_RegisterSerializationProvider.htm>`_
 - `IBsonSerializer <{+api-root+}/T_MongoDB_Bson_Serialization_IBsonSerializer.htm>`__
 - `SerializerBase<T> <{+api-root+}/T_MongoDB_Bson_Serialization_Serializers_SerializerBase_1.htm>`__
--- a/source/fundamentals/data-formats/serialization.txt
+++ b/source/fundamentals/data-formats/serialization.txt
@@ -18,6 +18,7 @@ serialization. Serialization is the process of mapping a C# object to a BSON
 document for storage in MongoDB.
 
 .. tip:: Serialization
+
    To learn more about serialization, see the :wikipedia:`Serialization </serialization>`
    article on Wikipedia.
 
@@ -35,10 +36,9 @@ For a full list of available serializers, see the
 Serializer Registry
 -------------------
 
-The serializer registry is an implementation of the ``IBsonSerializerRegistry`` interface
-that contains all registered serializers. Before you can use a serializer, you
-must register it by adding it to the serializer registry, as
-shown in the following example:
+The serializer registry contains all registered serializers that are available
+to your application. Before you can use a serializer, you must add it to the
+serializer registry, as shown in the following example:
 
 .. code-block:: csharp
 
--- a/source/fundamentals/data-formats/serialization.txt
+++ b/source/fundamentals/data-formats/serialization.txt
@@ -42,7 +42,7 @@ serializer registry, as shown in the following example:
 
 .. code-block:: csharp
 
-   BsonSerializer.RegisterSerializer(typeof(MySerializerType), new MySerializer); 
+   BsonSerializer.RegisterSerializer(new CustomTypeSerializer()); 
 
 To access the serializer registry, use the ``SerializerRegistry`` property
 of the ``BsonSerializer`` class as follows:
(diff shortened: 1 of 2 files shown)


source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+.. important::
+
+   The serializer registry is a global registry. This means that you cannot use
+   multiple registries in a single application.
+
+Serialization Provider
+----------------------
+
+A serialization provider allows you to provide serializers to your
+serialization registry. The serialization provider should be registered as early
```
mongoKart, line 40: I: I don't fully understand what it means to provide a serialization to the registry. Just to add it by calling a method?

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+as possible in your application to ensure that the serializers are used by the
+registry. Your serialization provider must implement the
```
mongoKart, line 42: I: Clarify "as early as possible"

mongoKart, line 42: I: Passive voice

```suggestion
as possible in your application to ensure that the registry uses your serializers.
Your serialization provider must implement the
```

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
+.. _csharp-serialization:
+
+=============
+Serialization
```
mongoKart, line 4: High-level suggestion after reading the whole page: Add a series of steps at the beginning of the page to describe the entire process (or something like that). The connection between the various moving pieces isn't always super clear.

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+A serialization provider allows you to provide serializers to your
+serialization registry. The serialization provider should be registered as early
+as possible in your application to ensure that the serializers are used by the
+registry. Your serialization provider must implement the
+``IBsonSerializationProvider`` interface. 
+
+Example
+~~~~~~~
+
+The following example implements a serialization provider. The provider has a
```
mongoKart, line 48: Q: Is it more accurate to say that the following class _is_ a serialization provider by virtue of its implementing the interface? As written, this seems to say that `IBsonSerializationProvider` is one of a number of serialization providers.

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+``GetSerializer`` method that returns a serializer based on the type of the
```
mongoKart, line 49:
```suggestion
The following example implements a serialization provider. The interface requires a
``GetSerializer`` method that returns a serializer based on the type of the
```

mongoKart, line 49:
```suggestion
The following example implements a serialization provider. The provider has a
``GetSerializer`` method that returns a serializer object based on the type of the
```

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+input. If the method input is an ``int`` it returns a custom ``Int32``
```
mongoKart, line 50:
```suggestion
input. If the method input is an ``int``, ``GetSerializer`` returns a custom ``Int32``
```

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+
+.. code-block:: csharp
+   
+   class MyProvider : IBsonSerializationProvider
+   {
+       public IBsonSerializer GetSerializer(Type type)
+       {
+           if (type == typeof(int))
+           {
+               return new MyInt32Serializer();
```
mongoKart, line 62: Q: I'm not sure how complex these custom serializer classes are. Would it be helpful to show how to define one on this page?

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+``IBsonSerializationProvider`` interface. 
+
+Example
+~~~~~~~
+
+The following example implements a serialization provider. The provider has a
+``GetSerializer`` method that returns a serializer based on the type of the
+input. If the method input is an ``int`` it returns a custom ``Int32``
+serializer, otherwise it returns ``null`` so that it can be handled by another
+provider.
```
mongoKart, line 52: I: comma splice

S:

```suggestion
serializer. If the input is any other type, the method returns ``null``, indicating that
another provider should return the appropriate serializer.
```

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+           {
+               return new MyInt32Serializer();
+           }        
+   
+           return null;
+       }
+   }
+
+You can register the serialization provider using the
+``RegisterSerializationProvider`` method as shown in the following example:
```
mongoKart, line 70: i: [using](https://www.mongodb.com/docs/meta/style-guide/terminology/alphabetical-terms/#std-term-using:~:text=word%2C%20username.-,using,-by%20using%20with)

S: 

```suggestion
You can use the ``RegisterSerializationProvider`` method to register the serialization provider,
as shown in the following example:
```

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+
+.. code-block:: csharp
+
+   BsonSerializer.RegisterSerializationProvider(new MyProvider);
+
+Serializers
```
mongoKart, line 76: It seems like this is something that should go earlier on the page, before the term "serializer" is used in other explanations

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+primitive types, collection types, and custom classes.
+
+For a full list of available serializers, see the
+`Serializers namespace API documentation <{+api-root+}/N_MongoDB_Bson_Serialization_Serializers.htm>`__.
+
+Custom Serializers
+~~~~~~~~~~~~~~~~~~
+
+In some cases, you may need to implement a custom serializer. When creating a
+custom serializer, your class should inherit from ``SerializerBase<T>`` and
```
mongoKart, line 91: I: Does "should" mean it's required or recommended?

S: If required, I'd cut it. If recommended, Style Guide suggests "We recommend" (or, honestly, just direct the user to do it as if it were required).

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+override the ``Deserialize()`` and ``Serialize()`` methods. 
+
+The following code example shows how to implement a custom serializer for
+an ``Int32``:
```
mongoKart, line 95: S: I assume that anyone reading this page knows C# well enough to understand that "int" is an alias for "Int32", but maybe not a bad idea to just use "int" everywhere.

```suggestion
an ``int``:
```

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+interface. The {+driver-short+} has many built-in serializers made to handle
+primitive types, collection types, and custom classes.
+
+For a full list of available serializers, see the
+`Serializers namespace API documentation <{+api-root+}/N_MongoDB_Bson_Serialization_Serializers.htm>`__.
+
+Custom Serializers
+~~~~~~~~~~~~~~~~~~
+
+In some cases, you may need to implement a custom serializer. When creating a
```
mongoKart, line 90: I: ["may" equals permission](https://www.mongodb.com/docs/meta/style-guide/terminology/alphabetical-terms/#std-term-may:~:text=than%20matrices.-,may,-Use%20may%20to)

```suggestion
In some cases, you might need to implement a custom serializer. When creating a
```

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+       public override void Serialize(BsonSerializationContext context, BsonSerializationArgs args, int value)
+       {
+           context.Writer.WriteInt32(value);
+       }
+   }
+
+.. important:: 
+
+   The preceding code example shows an overly simplistic serializer. When
+   creating a custom serializer ensure it handles all standard and edge cases.
```
mongoKart, line 115: S: Probably cleaner to avoid editorializing on the code sample. Maybe something like:

```suggestion
   When you create a custom serializer, you might need to implement more than 
   the ``Serialize`` and ``Deserialize`` methods to ensure it handles all
   standard and edge cases.
```

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+
+.. tip::
+
+   The {+driver-short+} provides a built-in `Int32Serializer 
+   <{+api-root+}/T_MongoDB_Bson_Serialization_Serializers_Int32Serializer.htm>`__.
+
+Opt-in Interfaces
```
mongoKart, line 122: good job on this capitalization. had to look it up to make sure it was right

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+-----------------
+
+The {+driver-short+} has several opt-in interfaces that you can implement with
```
mongoKart, line 125: S: What does it mean to be an "opt-in interface?" Probably helpful to clarify.

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+your custom serializer based on type of data your serializer handles.
```
mongoKart, line 126:
```suggestion
The {+driver-short+} has several opt-in interfaces that your custom serializer
class can implement, depending on the type of data the serializer handles.
```

source/fundamentals/data-formats/serialization.txt
mongoKart, line 49: S: The rest of our docs include parentheses in monospace when referring to a method. Makes sense to do the same thing here unless you have a reason not to. Applies throughout.

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,166 @@
...
+
+IBsonIdProvider
+~~~~~~~~~~~~~~~
+
+The `IBsonIdProvider
+<{+api-root+}/T_MongoDB_Bson_Serialization_IBsonIdProvider.htm>`__
+interface provides the ``GetDocumentId`` and ``SetDocumentId``
+methods, and is useful if your class uses an ID type other than ``ObjectId``.
```
mongoKart, line 134: q: "your class" = "the objects you intend to serialize with this class" ?

mongoKart (changes requested):
Suggested some restructuring and clarification. Going to be a really helpful page when it's done!

source/fundamentals/data-formats/serialization.txt
jordan-smith721: After some discussion with engineering, we determined that we should remove the serialization provider section since it's more of an internal mechanism that doesn't need to be surfaced to the user.

source/fundamentals/data-formats/serialization.txt
jordan-smith721: Removed serialization provider section

jordan-smith721: Removed serialization provider section

source/fundamentals/data-formats/serialization.txt
jordan-smith721: I restructured the page to have the serialization section first, that and removing the provider section makes it a little more clear I think. Let me know if you agree or think we still need to add some more clarification

source/fundamentals/data-formats/serialization.txt
jordan-smith721: Removed serialization provider section

source/fundamentals/data-formats/serialization.txt
jordan-smith721: Removed serialization provider section

jordan-smith721: Removed serialization provider section

source/fundamentals/data-formats/serialization.txt
jordan-smith721: Removed serialization provider section

source/fundamentals/data-formats/serialization.txt
jordan-smith721: Removed serialization provider section

source/fundamentals/data-formats/serialization.txt
jordan-smith721: Removed serialization provider section

source/fundamentals/data-formats/serialization.txt
jordan-smith721: Moved this section to the top of the page

source/fundamentals/data-formats/serialization.txt
jordan-smith721: It's recommended, but changed it to just direct the user to do it

source/fundamentals/data-formats/serialization.txt
jordan-smith721: After some more consideration I opted to remove this note. The intent was to make sure a user doesn't copy this thinking it would suffice as an Int32 serializer, but I think linking to the already existing Int32 serializer covers our bases there.

source/fundamentals/data-formats/serialization.txt
jordan-smith721: Added a bit of clarification, let me know if that works

source/fundamentals/data-formats/serialization.txt
jordan-smith721: Yes, updated the wording

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,127 @@
...
+that contains all registered serializers. Before you can use a serializer, you
+must register it by adding it to the serializer registry, as
+shown in the following example:
+
+.. code-block:: csharp
+
+   BsonSerializer.RegisterSerializer(typeof(MySerializerType), new MySerializer); 
+
+You can access the serializer registry with the ``SerializerRegistry`` property
+of the ``BsonSerializer`` class as follows:
```
mongoKart, line 45: I: [ambiguous "with"](https://www.mongodb.com/docs/meta/style-guide/terminology/alphabetical-terms/#std-term-using:~:text=word%2C%20username.-,using,-by%20using%20with)

```suggestion
To access the serializer registry, use the ``SerializerRegistry`` property
of the ``BsonSerializer`` class as follows:
```

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,127 @@
...
+
+.. important::
+
+   The serializer registry is a global registry. This means that you cannot use
+   multiple registries in a single application.
+
+Custom Serializers
+~~~~~~~~~~~~~~~~~~
+
+In some cases, you might need to create a custom serializer. When creating a
```
mongoKart, line 59: S: Does "in some cases" mean "if you need to serialize a custom type"? If so, probably clearer to just say that.

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,127 @@
...
+   class MyIntSerializer : SerializerBase<int>
+   {
+       public override int Deserialize(BsonDeserializationContext context, BsonDeserializationArgs args)
+       {
+           return context.Reader.ReadInt();
+       }
+   
+       public override void Serialize(BsonSerializationContext context, BsonSerializationArgs args, int value)
+       {
+           context.Writer.WriteInt(value);
```
mongoKart, line 77: S: It's a little confusing to demonstrate how to create a custom serializer an int, then say that the driver already includes one. Presumably a user would create a custom serializer only if the driver didn't already have one.

One option (probably harder but more helpful) is to show how to create a custom serializer for a custom type. Another option is to explain, maybe in the tip box, that the user wouldn't need to actually create one for ints because the driver includes it.

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,127 @@
...
+interface provides the ``GetDocumentId()`` and ``SetDocumentId()``
+methods, and is useful if the object you are serializing uses an ID type other than ``ObjectId``.
+
+IBsonDocumentSerializer
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Implementing the `IBsonDocumentSerializer
+<{+api-root+}/T_MongoDB_Bson_Serialization_IBsonDocumentSerializer.htm>`__
+interface enables the driver to access the member
+information of the object your are serializing. This allows the driver to
```
mongoKart, line 106:
```suggestion
information of the object you are serializing. This allows the driver to
```

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,127 @@
...
+
+.. tip::
+
+   The {+driver-short+} provides a built-in `Int32Serializer 
+   <{+api-root+}/T_MongoDB_Bson_Serialization_Serializers_Int32Serializer.htm>`__.
+
+Opt-in Interfaces
+-----------------
+
+The {+driver-short+} has several optional interfaces that your custom serializer
```
mongoKart, line 89: S: Are these the only three opt-in interfaces? If not, can we include a link to the rest?

mongoKart (changes requested):
Much clearer and easier to use. A few small things to change

source/fundamentals/data-formats/serialization.txt
jordan-smith721: I changed this up a bit. It's now a serializer for a custom object, but I removed the implementation of the different methods since there's not really any right or wrong way to actually implement them.

mongoKart (changes requested):
A couple small things

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,124 @@
...
+
+Overview
+--------
+
+In this guide, you can learn how to use the {+driver-long+} to perform
+serialization. Serialization is the process of mapping a C# object to a BSON
+document for storage in MongoDB.
+
+.. tip:: Serialization
+   To learn more about serialization, see the :wikipedia:`Serialization </serialization>`
```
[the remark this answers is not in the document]
jordan-smith721: Oops, fixed

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,124 @@
...
+primitive types, collection types, and custom classes.
+
+For a full list of available serializers, see the
+`Serializers namespace API documentation <{+api-root+}/N_MongoDB_Bson_Serialization_Serializers.htm>`__.
+
+Serializer Registry
+-------------------
+
+The serializer registry is an implementation of the ``IBsonSerializerRegistry`` interface
+that contains all registered serializers. Before you can use a serializer, you
```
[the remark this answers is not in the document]
jordan-smith721: Yeah I think you're right, I don't think they actually need to implement the interface so we're probably okay to remove it

mongoKart (approved):
LGTM

JamesKovacs (changes requested):
Looking good. We may want to consider giving additional guidance such as using `StructSerializerBase<T>` for scalar types. A more detailed example would also valuable that shows reading/writing to the `BsonSerializationContext`. Although the `MyInt32Serializer` is contrived, it demonstrates how you would actually leverage the context to read/write BSON. Let's brainstorm some ideas in Slack about a better example.

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,134 @@
...
+
+In some cases, you might need to create a custom serializer. When creating a
+custom serializer, implement the ``SerializerBase<T>`` abstract base class and
+override the ``Deserialize()`` and ``Serialize()`` methods.
+
+The following code example shows a custom ``BsonRegularExpression`` serializer:
+
+.. code-block:: csharp
+
+   class CustomRegularExpressionSerializer : SerializerBase<BsonRegularExpression>
```
JamesKovacs, line 70: This should be `Regex` rather than `BsonRegularExpression`. It isn't common to use BSON types (other than maybe `BsonDocument`) directly in your C# application. The serializer's job is to convert from database types (like `BsonRegularExpression`) to C# types like `Regex`.

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,134 @@
...
+   {
+       public override BsonRegularExpression Deserialize(BsonDeserializationContext context, BsonDeserializationArgs args)
```
JamesKovacs, line 72: BsonRegularExpression => Regex

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,134 @@
...
+     {
+         var type = context.Reader.GetCurrentBsonType();
+         switch (type)
+         {
+             case BsonType.RegularExpression:
+                 return context.Reader.ReadRegularExpression();
```
JamesKovacs, line 78: There is a helper `AsRegex` so you can write `return context.Reader.ReadRegularExpression().AsRegex;`

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,134 @@
...
+             case BsonType.String:
+                 var pattern = context.Reader.ReadString();
+                 return new Regex(pattern);
+             default:
+                 var message = string.Format("Cannot convert a {0} to a BsonRegularExpression.", type);
+                 throw new NotSupportedException(message);
```
JamesKovacs, line 84: You can write this more succinctly as:
```
throw new NotSupportedException($"Cannot convert a {type} to a BsonRegularExpression.");
```

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,134 @@
...
+         }
+     }
+
+     public override void Serialize(BsonSerializationContext context, BsonSerializationArgs args, BsonRegularExpression value)
```
JamesKovacs, line 88: BsonRegularExpression => Regex

JamesKovacs (changes requested):
Overall looking great. See comments inline.

source/fundamentals/data-formats/serialization.txt
```diff
@@ -0,0 +1,134 @@
...
+.. tip:: Serialization
+
+   To learn more about serialization, see the :wikipedia:`Serialization </serialization>`
+   article on Wikipedia.
+
+Serializers
+-----------
+
+Serializers are classes that handle the translation of C# objects to and
+from BSON documents. Most serializers implement the ``IBsonSerializer``
```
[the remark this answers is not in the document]
JamesKovacs: To be clear, I'm quibbling about "most". As far as I know, all of our serializers must implement `IBsonSerializer` either directly or via deriving from a base class that implements it.

JamesKovacs (approved):
LGTM

PR #195, opened 2024-05-10
DOCSP-37264: Add text comparing async vs sync/LINQ vs builder usecases

mcmorisi:
# Pull Request Info

JIRA - [DOCSP-37264](https://jira.mongodb.org/browse/DOCSP-37264)
Staging
* https://preview-mongodbmcmorisi.gatsbyjs.io/csharp/DOCSP-37264-new-sync-vs-async/faq/
* https://preview-mongodbmcmorisi.gatsbyjs.io/csharp/DOCSP-37264-new-sync-vs-async/usage-examples/

## Self-Review Checklist

### DIFF
--- a/source/faq.txt
+++ b/source/faq.txt
@@ -183,6 +183,14 @@ The error message consists of multiple parts:
      can use tools like ``openssl s_client`` to debug TLS/SSL-related
      certificate problems.
 
+.. _csharp-faq-linq-vs-builder:
+
+Should I Use LINQ or Builder Classes When Querying for Documents?
+-----------------------------------------------------------------
+
+If you define your documents using C# classes, use LINQ. If you use the
+``BsonDocument`` class to represent your documents, use builders. 
+
 .. _csharp-faq-unsupported-expressions:
 
 Why are Certain LINQ or Builder Expressions Unsupported?
--- a/source/faq.txt
+++ b/source/faq.txt
@@ -188,8 +188,12 @@ The error message consists of multiple parts:
 Should I Use LINQ or Builder Classes When Querying for Documents?
 -----------------------------------------------------------------
 
-If you define your documents using C# classes, use LINQ. If you use the
-``BsonDocument`` class to represent your documents, use builders. 
+If you're used to programming in C#, consider using LINQ because of its similar feel
+to programming in native C#. If you have prior experience with other MongoDB drivers,
+consider using Builder classes because of its consistency with other drivers.
+
+We encourage experimenting with both approaches to determine the most suitable framework
+for your purposes.
 
 .. _csharp-faq-unsupported-expressions:
 
--- a/source/faq.txt
+++ b/source/faq.txt
@@ -190,7 +190,7 @@ Should I Use LINQ or Builder Classes When Querying for Documents?
 
 If you're used to programming in C#, consider using LINQ because of its similar feel
 to programming in native C#. If you have prior experience with other MongoDB drivers,
-consider using Builder classes because of its consistency with other drivers.
+consider using Builder classes because of their consistency with other drivers.
 
 We encourage experimenting with both approaches to determine the most suitable framework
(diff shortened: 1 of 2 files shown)


source/faq.txt
```diff
@@ -183,6 +183,14 @@ The error message consists of multiple parts:
      can use tools like ``openssl s_client`` to debug TLS/SSL-related
      certificate problems.
 
+.. _csharp-faq-linq-vs-builder:
+
+Should I Use LINQ or Builder Classes When Querying for Documents?
+-----------------------------------------------------------------
+
+If you define your documents using C# classes, use LINQ. If you use the
+``BsonDocument`` class to represent your documents, use builders.
```
jordan-smith721, line 192: I see that this was the example used in the Jira ticket, but I don't think this is necessarily true. In our docs we define our documents with C# classes in both the LINQ and the Builders pages. I think the difference between LINQ and Builders is more based on your use case and the type of query you're building. We also decided when creating the C# docs that we wouldn't push users towards using BsonDocuments, but to advocate for POCOs instead in most/all cases.

I think we should check with the engineering team on whether we should add this comparison section, and if so, get their opinion on when to use one or the other.

source/usage-examples.txt
```diff
@@ -42,6 +42,16 @@ operations. Each example provides the following information:
 
 - The expected result after running the example
 
+.. tip::
+
+   Whether you use a synchronous or asynchronous framework in your application depends
+   on your use case. Synchronous calls are more suitable for simpler query workflows or
```
jordan-smith721, line 48:
```suggestion
   on your use case. Synchronous calls are more suitable for simple query workflows or
```

source/usage-examples.txt
```diff
@@ -42,6 +42,16 @@ operations. Each example provides the following information:
...
+   when you must implement sequential logic. Consider using asynchronous calls if
+   your application relies on multiple concurrent database requests or if your
+   program doesn't require an immediate response from the database to continue
+   executing. We encourage experimenting with both approaches to determine the most
```
jordan-smith721, line 52: Suggestion: Split this out for readability
```suggestion
   executing. 
   
   We encourage experimenting with both approaches to determine the most
```

jordan-smith721 (changes requested):
Nice work on this! A few suggestions and a comment on the builders vs LINQ section

source/faq.txt
mcmorisi: Good shout, will reach out to the C# team!

source/faq.txt
```diff
@@ -183,6 +183,18 @@ The error message consists of multiple parts:
      can use tools like ``openssl s_client`` to debug TLS/SSL-related
      certificate problems.
 
+.. _csharp-faq-linq-vs-builder:
+
+Should I Use LINQ or Builder Classes When Querying for Documents?
+-----------------------------------------------------------------
+
+If you're used to programming in C#, consider using LINQ because of its similar feel
+to programming in native C#. If you have prior experience with other MongoDB drivers,
```
jordan-smith721, line 192:
```suggestion
If you're used to programming in {+language+}, consider using LINQ because of its similar feel
to programming in native {+language+}. If you have prior experience with other MongoDB drivers,
```

source/faq.txt
```diff
@@ -183,6 +183,18 @@ The error message consists of multiple parts:
...
+consider using Builder classes because of their consistency with other drivers.
+
+We encourage experimenting with both approaches to determine the most suitable framework
```
jordan-smith721, line 195: Suggestion: use something other than "framework" here since they aren't technically frameworks (I'm not set on mechanism either but it was the best word I could think of at the moment)

```suggestion
We encourage experimenting with both approaches to determine the most suitable mechanism
```

jordan-smith721 (approved):
LGTM w/ a couple suggestions

PR #208, opened 2024-06-07
DOCSP-33408: Add $vectorSearch documentation

mcmorisi:
# Pull Request Info

Pardon the long line with the vector definition, the example is taken directly from the Atlas manual.

JIRA - [DOCSP-33408](https://jira.mongodb.org/browse/DOCSP-33408)
Staging - 

* <https://preview-mongodbmcmorisi.gatsbyjs.io/csharp/DOCSP-33408-vectorsearch/fundamentals/linq/>
* <https://preview-mongodbmcmorisi.gatsbyjs.io/csharp/DOCSP-33408-vectorsearch/fundamentals/builders/>

## Self-Review Checklist


source/fundamentals/linq.txt
```diff
@@ -497,6 +497,82 @@ The following is a subset of the returned results:
        ]
     }
 
+$vectorSearch
+~~~~~~~~~~~~~
+
+The ``$vectorSearch`` aggregation stage performs an Approximate Nearest Neighbor search
```
mongoKart, line 503: S: It looks like Approximate Nearest Neighbor isn't usually capitalized. [SG recommends italics for new terms](https://www.mongodb.com/docs/meta/style-guide/style/text-formatting/#:~:text=HTTP%20check%20box.-,Terms%2C%20new,-Format%20using%20italics)

```suggestion
The ``$vectorSearch`` aggregation stage performs an *approximate nearest neighbor* search
```

source/fundamentals/linq.txt
```diff
@@ -497,6 +497,82 @@ The following is a subset of the returned results:
...
+on an Atlas Vector Search index in the specified field.
```
mongoKart, line 504: S: I think it's more common to search "on" a field than "in", unless you're searching within a single value of a single field for a substring or something

```suggestion
on an Atlas Vector Search index on the specified field.
```

source/fundamentals/linq.txt
```diff
@@ -497,6 +497,82 @@ The following is a subset of the returned results:
...
...157,-0.010469886,0.0016852784,0.01716768,-0.008990373,-0.001987679,0.026417969,0.023792166,0.0046917885,-0.0071909656,-0.00032051947,-0.023259008,-0.009170313,0.02071318,-0.03156294,-0.030869836,-0.006324584,0.013795458,-0.00047151142,0.016874444,0.00947688,0.00985009,-0.029883493,0.024205362,-0.013522214,-0.015075036,-0.030603256,0.029270362,0.010503208,0.021539574,0.01743426,-0.023898797,0.022019416,-0.0068777353,0.027857494,-0.021259667,0.0025758184,0.006197959,0.006447877,-0.00025200035,-0.004941706,-0.021246338,-0.005504854,-0.008390571,-0.0097301295,0.027244363,-0.04446536,0.05216949,0.010243294,-0.016008062,0.0122493,-0.0199401,0.009077012,0.019753495,0.006431216,-0.037960835,-0.027377652,0.016381273,-0.0038620618,0.022512587,-0.010996379,-0.0015211658,-0.0102233,0.007071005,0.008230623,-0.009490209,-0.010083347,0.024431955,0.002427534,0.02828402,0.0035721571,-0.022192692,-0.011882754,0.010056688,0.0011904413,-0.01426197,-0.017500903,-0.00010985966,0.005591492,-0.0077707744,-0.012049366,0.011869425,0.00858384,-0.024698535,-0.030283362,0.020140035,0.011949399,-0.013968734,0.042732596,-0.011649498,-0.011982721,-0.016967745,-0.0060913274,-0.007130985,-0.013109017,-0.009710136};
```
mongoKart, line 540: lmao

is this the best option here (genuinely asking)?

source/fundamentals/linq.txt
```diff
@@ -497,6 +497,82 @@ The following is a subset of the returned results:
...
+      public string Title { get; set; }
+
+      public string Plot { get; set; }
+
+      [BsonElement("plot_embedding")]
+      public double[] Embedding { get; set; }
+   }
+
+The following example shows how to generate a ``$vectorSearch`` stage using LINQ to search
+the ``plot_embedding`` field using vector embeddings for the string ``"time travel"``:
```
mongoKart, line 535: Does the vectorSearch stage use LINQ? "The following example shows how to generate a ``$vectorSearch`` stage that uses LINQ..."

 Or does the example use LINQ to generate a vectorSearch stage? "The following example shows how to use LINQ to generate a $vectorSearch stage..."

source/fundamentals/linq.txt
```diff
@@ -497,6 +497,82 @@ The following is a subset of the returned results:
...
+   { "_id" : ObjectId("573a13c6f29313caabd715d3"), "plot" : "Agent J travels in time to M.I.B.'s early days in 1969 to stop an alien from assassinating his friend Agent K and changing history.", "title" : "Men in Black 3" }
+   { "_id" : ObjectId("573a13d4f29313caabd98c13"), "plot" : "Bound by a shared destiny, a teen bursting with scientific curiosity and a former boy-genius inventor embark on a mission to unearth the secrets of a place somewhere in time and space that exists in their collective memory.", "title" : "Tomorrowland" }
+   { "_id" : ObjectId("573a13b6f29313caabd477fa"), "plot" : "With the help of his uncle, a man travels to the future to try and bring his girlfriend back to life.", "title" : "Love Story 2050" }
+   { "_id" : ObjectId("573a13e5f29313caabdc40c9"), "plot" : "A dimension-traveling wizard gets stuck in the 21st century because cell-phone radiation interferes with his magic. With his home world on the brink of war, he seeks help from a jaded ...", "title" : "The Portal" }
+
+For more information about the Atlas Vector Search, Atlas Vector Search indexes, and how
```
mongoKart, line 570:
```suggestion
For more information about Atlas Vector Search, Atlas Vector Search indexes, and how
```

mongoKart (approved):
LGTM. Really nice job. A few suggestions and small things to clarify

source/fundamentals/linq.txt
mcmorisi: It's the latter, can update the wording to clarify

source/fundamentals/linq.txt
mcmorisi: So this only prints on one line that just goes way off to the right – it's not super important to have it from a tutorial standpoint but it's helpful if people want to copy in and run the sample themselves.

source/fundamentals/linq.txt
```diff
@@ -497,6 +497,82 @@ The following is a subset of the returned results:
        ]
     }
 
+$vectorSearch
+~~~~~~~~~~~~~
+
+The ``$vectorSearch`` aggregation stage performs an *approximate nearest neighbor* search
+on an Atlas Vector Search index on the specified field.
```
BorisDog, line 504: I think "peforms ... search on a vector in the specified field." would be more accurate?
Like [here](https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-stage/#mongodb-pipeline-pipe.-vectorSearch).
Also worth specify the need for vector index explicitly (besides Tip)?

source/fundamentals/linq.txt
```diff
@@ -497,6 +497,82 @@ The following is a subset of the returned results:
...
+      [BsonIgnoreIfDefault]
+      public string Title { get; set; }
+
+      public string Plot { get; set; }
+
+      [BsonElement("plot_embedding")]
+      public double[] Embedding { get; set; }
+   }
+
+The following example shows how to use LINQ to generate a ``$vectorSearch`` stage to search
```
BorisDog, line 534: This is not really LINQ.  It's just `IMongoQueryable` is the entry point for VectorSearch stage, as well as for LINQ expressions. But the next `Select` statement is part of LINQ..

source/fundamentals/linq.txt
```diff
@@ -497,6 +497,82 @@ The following is a subset of the returned results:
        ]
     }
 
+$vectorSearch
```
BorisDog, line 500: This part looks good. 
VectorSearch is not only limited to IMongoQueryable. We also have [builders](https://github.com/mongodb/mongo-csharp-driver/blob/master/tests/MongoDB.Driver.Tests/PipelineDefinitionBuilderTests.cs#L411C11-L411C12) and [Fluent API](https://github.com/mongodb/mongo-csharp-driver/blob/master/tests/MongoDB.Driver.Tests/Search/AtlasSearchTests.cs#L577) options. I think it's worth documenting all options.

Also should we have a separate page for `VectorSearch` (like we do for `AtlasSearch`) to increase visibility? cc @rishitb-mongodb?

BorisDog (changes requested):

source/fundamentals/linq.txt
mcmorisi: I was curious about this point and would like to hear Rishit's PoV – I originally didn't add documentation to the Builders page we have since it didn't seem that there was a specific builder for the VectorSearch stage (i.e. nothing like a  

I'd also want to be wary of duplicating information – there is ample documentation on Vector Search in Atlas already so I'd prefer to link to that when able, unless there are C# driver-specific quirks that are worth calling out.

mcmorisi: I guess I'm struggling with the "natural" point to describe Vector Search in our docs since it seems like neither Builders nor LINQ is fully appropriate. Maybe a section in the Aggregations page with a link to the detailed documentation?

rishitb-mongodb: I think we need to have some sort of consistency, either within our documentation (by having a dedicated page for Atlas Vector Search just like we do for Atlas Search) OR with other driver's documentation (which is what the above approach does by mimicing what the Java driver was doing for Atlas Vector Search). I know there is a standardization effort coming soon for this. I didn't know that we have Builders and FluentAPI option for Vector Search as well. If this is the case then I agree this page shouldn't be under LINQ.  If a developer can do Vector Search using Builders, Fluent API or LINQ then we should have a dedicated page for it but we don't need to repeat everything in the main Atlas documentation, just redirect to it wherever needed.

mongoKart (approved):
LGTM

source/fundamentals/builders.txt
```diff
@@ -347,6 +347,68 @@ pipeline stage.
...
+
+   // Specifies that the vector search will consider the 150 nearest neighbors
+   // in the specified index
+   var options = new VectorSearchOptions<EmbeddedMovie>() {
```
BorisDog, line 381: Please switch to Alman style braces, here and in builders.

BorisDog (changes requested):

PR #217, opened 2024-07-18
Clarify using commas in connection strings

stephmarie17:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-41035>
Staging - <https://preview-mongodbstephmarie17.gatsbyjs.io/csharp/docsp-41035-updateConnectionString/fundamentals/enterprise-authentication/#mongodb-oidc>

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/enterprise-authentication.txt
+++ b/source/fundamentals/enterprise-authentication.txt
@@ -298,7 +298,7 @@ see the corresponding syntax.
       Replace the ``<percent-encoded audience>`` placeholder with the percent-encoded
       value of the ``audience`` parameter configured on your MongoDB deployment. 
 
-      You can't pass values containing the comma (``,``) character to the ``authMechanismProperties`` 
+      You cannot pass values containing the comma (``,``) character to the ``authMechanismProperties`` 
       option. You must specify values that contain commas in a ``MongoCredential`` object, 
       as demonstrated in the :guilabel:`MongoCredential` tab.
       
@@ -352,7 +352,7 @@ see the corresponding syntax.
       Replace the ``<audience>`` placeholder with the
       value of the ``audience`` parameter configured on your MongoDB deployment.
 
-      You can't pass values containing the comma (``,``) character to the ``authMechanismProperties`` 
+      You cannot pass values containing the comma (``,``) character to the ``authMechanismProperties`` 
       option. You must specify values that contain commas in a ``MongoCredential`` object, 
       as demonstrated in the :guilabel:`MongoCredential` tab.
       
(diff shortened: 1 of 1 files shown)


source/fundamentals/enterprise-authentication.txt
```diff
@@ -297,6 +297,10 @@ see the corresponding syntax.
       The following code example shows how to specify Azure IMDS OIDC authentication. 
       Replace the ``<percent-encoded audience>`` placeholder with the percent-encoded
       value of the ``audience`` parameter configured on your MongoDB deployment. 
+
+      You can't pass values containing the comma (``,``) character to the ``authMechanismProperties``
```
stephmarie17, line 301: [note to reviewer] I noticed in our [style guide](https://www.mongodb.com/docs/meta/style-guide/writing/use-positive-statements/) we recommend using positive statements, but I thought this wording was clearer than saying something like "Avoid using...". However, I'm open to suggestions here!

jordan-smith721, line 301: Small suggestion: Although contractions are allowed by our style guide, it's also [noted here](https://www.mongodb.com/docs/meta/style-guide/style/contractions/#contractions) that if you want to emphasize the negative, don't use contractions. I think here we want to emphasize the negative
```suggestion
      You cannot pass values containing the comma (``,``) character to the ``authMechanismProperties`` 
```

jordan-smith721: Agreed, I think this is more clear than saying something like "Avoid using..."

source/fundamentals/enterprise-authentication.txt
```diff
@@ -346,7 +350,11 @@ see the corresponding syntax.
       The following code example shows how to specify GCP IMDS OIDC authentication as
       part of the authentication string. 
       Replace the ``<audience>`` placeholder with the
-      value of the ``audience`` parameter configured on your MongoDB deployment. 
+      value of the ``audience`` parameter configured on your MongoDB deployment.
+
+      You can't pass values containing the comma (``,``) character to the ``authMechanismProperties`` 
+      option. You must specify values that contain commas in a ``MongoCredential`` object, 
+      as demonstrated in the :guilabel:`MongoCredential` tab.
```
jordan-smith721, line 357: S: I wonder if these should be in `.. note::` admonitions, since I would assume this is a smaller edge-case and it seems to be a sort of side-comment before the example. That being said, part of me is also in favor of less admonitions, so I can be swayed either way. I'll leave it to your discretion, just wanted to note my thoughts on it!

jordan-smith721 (approved):
LGTM w/ a couple suggestions!

source/fundamentals/enterprise-authentication.txt
stephmarie17: That makes sense, will update!

source/fundamentals/enterprise-authentication.txt
stephmarie17: Ah I see what you mean. In the ticket, there was a link to the [Java docs](https://www.mongodb.com/docs/drivers/java/sync/current/fundamentals/enterprise-auth/#mongodb-oidc) as an example that have a similar caveat in the description, so I followed that pattern here instead of following the `.. note::` route.

PR #231, opened 2024-09-17
DOCSP-43125 - Remove SDAM Logging method

mongoKart:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-43125

### Staging Links

<li><a href=https://deploy-preview-231--mongodb-docs-csharp.netlify.app/upgrade>upgrade</a></li>

## Self-Review Checklist


jordan-smith721 (approved):
LGTM!

source/upgrade.txt
```diff
@@ -66,6 +66,11 @@ Version 3.0 Potential Breaking Change
   configuring authentication in the {+driver-short+}, see
   :ref:`Authentication Mechanisms. <csharp-authentication-mechanisms>`
 
+- This version removes the ``ClusterBuilder.ConfigureSdamLogging()`` method.
+  To configure SDAM logging,
```
BorisDog, line 70: It might sound that the new configuration is specific for SDAM.
Maybe it should be "SDAM or any other component" and point to our logging page https://www.mongodb.com/docs/drivers/csharp/current/fundamentals/logging/?

mongoKart: When would you use `MongoClientSettings.LoggingSettings` vs. `ClusterBuilder.ConfigureLoggingSettings`? It looks like they both take an instance of `LoggingSettings`.

rstam: `ClusterBuilder` is an internal component. Users shouldn't be using it.

rstam (changes requested):

source/upgrade.txt
BorisDog: Exactly, some more context: 
Unfortunately it is still a public API, and we've been recommending configuring SDAM logging before we had a proper logging mechanism.
So we still need to mention this potentially breaking change and point to our regular logging.

BorisDog (approved):
LGTM

PR #242, opened 2024-09-26
DOCSP-43911 - Implicit Client-Side Projection

mongoKart:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-43911

### Staging Links

<li><a href=https://deploy-preview-242--mongodb-docs-csharp.netlify.app/whats-new>whats-new</a></li>

## Self-Review Checklist

### DIFF
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -57,8 +57,38 @@ The 3.0 driver release includes the following new features:
   This type is available in .NET 5 and later.
   To learn more about the ``Half`` type, see the
   `Half Struct <https://learn.microsoft.com/en-us/dotnet/api/system.half?view=net-8.0>`__
-  API reference page on MSDN. 
+  API reference page on MSDN.
 
+- Adds support for implicit client-side projection when using the ``Find()`` and
+  ``Select()`` methods. In previous versions of the driver, you could perform client-side
+  projection only after calling the ``ToEnumerable()`` or ``AsEnumerable()`` method, and
+  only by using the ``Select()`` method, similar to the following examples:
+
+  .. code-block:: csharp
+
+     var find = collection
+         .Find(doc => doc.Id == 1)
+         .ToEnumerable()
+         .Select(doc => new { R = MyFunction(doc.Name) });
+  
+     var enumerable = collection.AsQueryable()
+         .Where(doc => doc.Id == 1)
+         .AsEnumerable()
+         .Select(doc => new { R = MyFunction(doc.Name) });
+
+  In {+driver-short+} v3.0, you can perform client-side projection directly in the
+  ``Find()`` and ``Select()`` methods, as shown in the following examples:
+
+  .. code-block:: csharp
+
+     var find = collection
+         .Find(doc => doc.Id == 1)
+         .Project(doc => new { R = MyFunction(doc.Name) });
+
+     var queryable = collection.AsQueryable()
+         .Where(doc => doc.Id == 1)
+         .Select(doc => new { R = MyFunction(doc.Name) });
+ 
 .. _csharp-version-2.28:
 
 What's New in 2.28
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -86,7 +86,7 @@ The 3.0 driver release includes the following new features:
   the ``Select()`` method, or the ``Project()`` aggregation stage with the LINQ3 provider.
   In previous versions of the driver, you could perform client-side
   projection with the LINQ3 provider only after calling the ``ToEnumerable()`` or
-  ``AsEnumerable()`` method, and only within the ``Select()`` method.
+  ``AsEnumerable()`` method.
   
   To learn how to enable and use client-side projection for a driver method, select the
   corresponding tab:
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -71,36 +71,76 @@ The 3.0 driver release includes the following new features:
   `TimeOnly Struct. <https://learn.microsoft.com/en-us/dotnet/api/system.timeonly?view=net-6.0>`__
   API reference page on MSDN.
 
-- Adds support for implicit client-side projection when using the ``Find()`` and
-  ``Select()`` methods. In previous versions of the driver, you could perform client-side
+- Adds support for client-side projection when using the ``Find()`` method,
+  the ``Select()`` method, or the ``Project()`` aggregation stage.
+  In previous versions of the driver, you could perform client-side
   projection only after calling the ``ToEnumerable()`` or ``AsEnumerable()`` method, and
-  only by using the ``Select()`` method, similar to the following examples:
-
-  .. code-block:: csharp
-
-     var find = collection
-         .Find(doc => doc.Id == 1)
-         .ToEnumerable()
-         .Select(doc => new { R = MyFunction(doc.Name) });
+  only within the ``Select()`` method.
   
-     var enumerable = collection.AsQueryable()
-         .Where(doc => doc.Id == 1)
-         .AsEnumerable()
-         .Select(doc => new { R = MyFunction(doc.Name) });
+  Select a tab below to see how to enable and use client-side projection for the
+  corresponding method.
 
-  In {+driver-short+} v3.0, you can perform client-side projection directly in the
-  ``Find()`` and ``Select()`` methods, as shown in the following examples:
+  .. tabs::
 
-  .. code-block:: csharp
+    .. tab:: Find() Method
+       :tabid: find-method
 
-     var find = collection
-         .Find(doc => doc.Id == 1)
-         .Project(doc => new { R = MyFunction(doc.Name) });
+       .. code-block:: csharp
 
-     var queryable = collection.AsQueryable()
-         .Where(doc => doc.Id == 1)
-         .Select(doc => new { R = MyFunction(doc.Name) });
- 
+          // Enable client-side projection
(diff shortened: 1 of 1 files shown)


jordan-smith721 (approved):
LGTM

source/whats-new.txt
```diff
@@ -57,8 +57,38 @@ The 3.0 driver release includes the following new features:
   This type is available in .NET 5 and later.
   To learn more about the ``Half`` type, see the
   `Half Struct <https://learn.microsoft.com/en-us/dotnet/api/system.half?view=net-8.0>`__
-  API reference page on MSDN. 
-
+  API reference page on MSDN.
+
+- Adds support for implicit client-side projection when using the ``Find()`` and
```
rstam, line 62: It won't be implicit.

Client-side projections are potentially undesirable and the current thinking is that they will have to "enable" client side projections in order to use them.

The current PR is not merged to master yet, but the two places client-side projections can be enabled are in the `MongoClientSettings` or in the `FindOptions` or `AggregateOptions` for an individual query.

Theoretically they could be implicit, but that's not the current thinking.

source/whats-new.txt
```diff
@@ -57,8 +57,38 @@ The 3.0 driver release includes the following new features:
...
+  ``Select()`` methods. In previous versions of the driver, you could perform client-side
```
rstam, line 63: Beside the `Find` and `Select` methods this will apply to the `Project` pipeline stage builder (which is analogous to the `Select` method).

source/whats-new.txt
```diff
@@ -57,8 +57,38 @@ The 3.0 driver release includes the following new features:
...
+  projection only after calling the ``ToEnumerable()`` or ``AsEnumerable()`` method, and
+  only by using the ``Select()`` method, similar to the following examples:
+
+  .. code-block:: csharp
+
+     var find = collection
+         .Find(doc => doc.Id == 1)
+         .ToEnumerable()
```
rstam, line 71: This is not driver supported client-side projections. This is the user doing an EXPLICIT client-side projection by splitting their query into server-side and client-side parts and separating them by `ToEnumerable`.

source/whats-new.txt
```diff
@@ -57,8 +57,38 @@ The 3.0 driver release includes the following new features:
...
+         .Select(doc => new { R = MyFunction(doc.Name) });
+  
+     var enumerable = collection.AsQueryable()
+         .Where(doc => doc.Id == 1)
+         .AsEnumerable()
```
rstam, line 76: Also EXPLICIT like the `Find` example above.

source/whats-new.txt
```diff
@@ -57,8 +57,38 @@ The 3.0 driver release includes the following new features:
...
+         .Select(doc => new { R = MyFunction(doc.Name) });
+
+  In {+driver-short+} v3.0, you can perform client-side projection directly in the
+  ``Find()`` and ``Select()`` methods, as shown in the following examples:
```
rstam, line 80: Yes but... they have to enable client-side projections to do this.

rstam (changes requested):

source/whats-new.txt
mongoKart: My understanding was that this is _currently_ the only way to do it, which is why I presented it as the "before" example. Is that wrong?

Either way, I rearranged a bit to hopefully make it clearer.

source/whats-new.txt
mongoKart: Thanks for clarifying! I was trying to figure this out from the relevant ticket and PR but must have missed this aspect.

source/whats-new.txt
```diff
@@ -82,6 +82,76 @@ The 3.0 driver release includes the following new features:
...
+
+          // Enable client-side projection
+          var aggregateOptions = new AggregateOptions;
+          aggregateOptions.TranslationOptions = new ExpressionTranslationOptions
+          {
+              EnableClientSideProjections = true
+          };
+
+          var aggregate = collection
+              .Aggregate(pipeline, aggregateOptions)
```
mongoKart, line 141: Some of your tests call `Aggregate()` with no arguments, but I couldn't find that method overload and wasn't sure whether it accepted an `options` argument--hence the `pipeline` argument.

source/whats-new.txt
```diff
@@ -82,6 +82,76 @@ The 3.0 driver release includes the following new features:
...
+  .. tip:: MongoClientSettings
+
+     To enable client-side projection for all queries on a client, set the
+     ``TranslationOptions`` property of the ``MongoClientSettings`` object to ``true``.
+  
+  To learn more about using the aggregation pipeline with the {+driver-short+}, see
+  :ref:`csharp-aggregation`.
+
+For more information about this release, see the :github:`v3.0 release notes
+</mongodb/mongo-csharp-driver/releases/tag/v3.0.0>`.
```
mongoKart, line 153: Link will be broken until actual release.

source/whats-new.txt
rstam: The correct syntax is:

```collection.Aggregate(aggregateOptions)```

There is no overload that takes a `pipeline`.

`Aggregate` is an *extension* method:

``` public static IAggregateFluent<TDocument> Aggregate<TDocument>(this IMongoCollection<TDocument> collection, AggregateOptions options = null)```

and the `options` argument is optional so:

```collection.Aggregate()```

is the same as:

```collection.Aggregate(options: null)```

File: source/whats-new.txt [hunk 8]
```diff
@@ -82,6 +82,76 @@ The 3.0 driver release includes the following new features:
   `TimeOnly Struct. <https://learn.microsoft.com/en-us/dotnet/api/system.timeonly?view=net-6.0>`__
   API reference page on MSDN.
 
+- Adds support for client-side projection when using the ``Find()`` method,
+  the ``Select()`` method, or the ``Project()`` aggregation stage.
+  In previous versions of the driver, you could perform client-side
+  projection only after calling the ``ToEnumerable()`` or ``AsEnumerable()`` method, and
+  only within the ``Select()`` method.
```
rstam, line 89: In previous versions of the driver behavior depended on which LINQ provider you were using:

* LINQ2: implicit client-side projections in `Find`, `Project` and `Select`
* LINQ3: explicit client-side projections only

In the 3.0 version of the driver implicit client-side projections are supported but only if `EnableClientSideProjections == true`

By "explicit client-side projections" I mean calling `ToEnumerable()` or `AsEnumerable()` to specify where the client-side part of the query starts.

source/whats-new.txt
```diff
@@ -82,6 +82,76 @@ The 3.0 driver release includes the following new features:
...
+
+  .. tabs::
+
+    .. tab:: Find() Method
+       :tabid: find-method
+
+       .. code-block:: csharp
+
+          // Enable client-side projection
+          var findOptions = new FindOptions;
```
rstam, line 102: var findOptions = new FindOptions();

missing parenthesis

source/whats-new.txt
```diff
@@ -82,6 +82,76 @@ The 3.0 driver release includes the following new features:
...
+       .. code-block:: csharp
+
+          // Enable client-side projection
+          var translationOptions = new ExpressionTranslationOptions
+          {
+              EnableClientSideProjections = true
+          };
+         
+          var queryable = collection
+              .AsQueryable(translationOptions)
```
rstam, line 124: This is subtle but the actual argument to `AsQueryable` is an `AggregateOptions` value, not an `ExpressionTranslationOptions` value.

We might want to be explicit about that.

`AsQueryable(translationOptions)` does work but only because there is an implicit conversion from `ExpressionTranslationOptions` to `AggregateOptions`, and I'm not sure we're going to keep that implicit conversion.

rstam (changes requested):

source/whats-new.txt
mongoKart: Thanks. I reworded this section. Is it accurate to say that v3.0 supports _implicit_ client-side projection when using the LINQ3 provider, so long as that setting is true?

source/whats-new.txt
mongoKart: Maybe I confused it with [Aggregate<TResult>](https://mongodb.github.io/mongo-csharp-driver/2.29.0/api/MongoDB.Driver/MongoDB.Driver.IMongoCollection-1.Aggregate.html)

source/whats-new.txt
```diff
@@ -82,6 +82,77 @@ The 3.0 driver release includes the following new features:
   `TimeOnly Struct. <https://learn.microsoft.com/en-us/dotnet/api/system.timeonly?view=net-6.0>`__
   API reference page on MSDN.
 
+- Adds support for implicit client-side projection when using the ``Find()`` method,
+  the ``Select()`` method, or the ``Project()`` aggregation stage with the LINQ3 provider.
+  In previous versions of the driver, you could perform client-side
+  projection with the LINQ3 provider only after calling the ``ToEnumerable()`` or
+  ``AsEnumerable()`` method, and only within the ``Select()`` method.
```
rstam, line 89: > and only within the ``Select()`` method

Why would this be the case? Once you're client-side you can do anything.

rstam (changes requested):
Sorry one small question I should have asked in a previous review.

source/whats-new.txt
mongoKart: I suspect I inferred this from something in the Jira ticket.

rstam (approved):
LGTM

PR #253, opened 2024-10-14
DOCSP-34044: Add bitwise operators to aggregation pipeline

mayaraman19:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-34044>

### Staging Links

<li><a href=https://deploy-preview-253--mongodb-docs-csharp.netlify.app/fundamentals/linq>fundamentals/linq</a></li>

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/linq.txt
+++ b/source/fundamentals/linq.txt
@@ -598,8 +598,7 @@ stage. The following guidelines apply when using bitwise operators:
 
 - All operands must be of type ``int`` or ``long``.
 
-- ``$bitAnd``, ``$bitOr``, and ``$bitXor`` all take two or more operands. 
-  ``$bitNot`` can only take one operand.
+- ``$bitAnd``, ``$bitOr``, and ``$bitXor`` all take two or more operands. ``$bitNot`` can only take one operand.
 
 - The bitwise operation will be evaluated from left to right.
 
--- a/source/fundamentals/linq.txt
+++ b/source/fundamentals/linq.txt
@@ -70,9 +70,9 @@ as follows:
 .. code-block:: csharp
    :emphasize-lines: 3
 
-    var restaurantsDatabase = client.GetDatabase("sample_restaurants");
-    var restaurantsCollection = restaurantsDatabase.GetCollection<Restaurant>("restaurants");
-    var queryableCollection = restaurantsCollection.AsQueryable();
+   var restaurantsDatabase = client.GetDatabase("sample_restaurants");
+   var restaurantsCollection = restaurantsDatabase.GetCollection<Restaurant>("restaurants");
+   var queryableCollection = restaurantsCollection.AsQueryable();
 
 The ``AsQueryable()`` method returns an `IMongoQueryable
 <{+new-api-root+}/MongoDB.Driver/MongoDB.Driver.Linq.IMongoQueryable.html>`__ instance that
@@ -591,11 +591,16 @@ in the Atlas manual and select :guilabel:`C#` from the language dropdown.
 Bitwise Operators
 ~~~~~~~~~~~~~~~~~
 
-This version of the {+driver-short+} supports the following bitwise operators in 
-the aggregation pipeline. You can use multiple bitwise operators within the same 
-stage. All operands must be of type ``int`` or ``long``. ``$bitAnd``, ``$bitOr``,
+The {+driver-short+} supports the use of the following 
+`bitwise operators <https://en.wikipedia.org/wiki/Bitwise_operation>`__ in the 
+aggregation pipeline. You can use multiple bitwise operators in the same 
+stage. The following limits apply when using bitwise operators:
+
+All operands must be of type ``int`` or ``long``. ``$bitAnd``, ``$bitOr``,
 and ``$bitXor`` all take two or more operands. ``$bitNot`` can only take one
 operand.
+
+The bitwise operation will be evaluated from left to right.
 
 The examples for this section use the following documents in a collection called
 ``ingredients``:
@@ -625,36 +630,35 @@ collection:
       public int Count { get; set; }
    }
 
-.. note::
+.. note:: Missing or Undefined Operands
 
-   For all bitwise operators, if the operand list contains a missing or undefined
-   value, the entire expression evaluates to ``null``.
-
+   If the operand list you pass to any bitwise operator contains a missing or 
+   undefined value, the entire expression evaluates to ``null``.
 
 $bitAnd
 +++++++
 
-The ``$bitAnd`` aggregation stage performs a bitwise AND operation on the given
-arguments. The following example shows how to generate a ``$bitAnd`` stage using
-LINQ:
+The ``$bitAnd`` aggregation operator performs a bitwise AND operation on the given
+arguments. You can use the ``$bitAnd`` operator by connecting two or more 
+clauses with a ``&`` character. The following example shows how to create a 
+``$bitAnd`` stage using LINQ: 
 
 .. code-block:: csharp
 
-    var query = queryableCollection.AsQueryable()
+    var query = queryableCollection
       .Where(i => i.Name == "eggs")
       .Select(i => i.Price & i.Count);
 
-The preceding example finds the document where the ``Name`` field has the value
-``"eggs"``. It then performs a bitwise AND operation on the values of the 
-``Price`` and ``Count`` fields in this document. The result contains one value:
-``4``.
+The preceding example retrieves the document in which the ``Name`` field has the 
+value ``"eggs"``. It then performs a bitwise AND operation on the values of the 
+``Price`` and ``Count`` fields in this document, returning a value of ``4``.
 
 The following example performs the same bitwise AND operation, but on all 
 documents in the collection:
 
 .. code-block:: csharp
 
-    var query = queryableCollection.AsQueryable()
+    var query = queryableCollection
       .Select(i => i.Price & i.Count);
 
 The result of the preceding example contains the following values:
@@ -675,30 +679,33 @@ the expression evaluates to ``null``.
 $bitOr
 ++++++
 
-The ``$bitOr`` aggregation stage performs a bitwise OR operation on the given
-arguments. The following example shows how to generate a ``$bitOr`` stage using
-LINQ:
+The ``$bitOr`` aggregation operator performs a bitwise OR operation on the given
+arguments. You can use the ``$bitOr`` operator by connecting two or more 
+clauses with a ``|`` character. The following example shows how to create a 
+``$bitOr`` stage using LINQ:
 
 .. code-block:: csharp
 
-    var query = queryableCollection.AsQueryable()
+    var query = queryableCollection
       .Where(i => i.Name == "eggs")
       .Select(i => i.Price | i.Count);
 
-The preceding example finds the document where the ``Name`` field has the value
-``"eggs"``. It then performs a bitwise OR operation on the values of the 
-``Price`` and ``Count`` fields in this document. The result contains one value:
+The preceding example retrieves the document in which the ``Name`` field has the 
+value ``"eggs"``. It then performs a bitwise OR operation on the values of the 
+``Price`` and ``Count`` fields in this document, returning the value
 ``13``.
 
 $bitNot
 +++++++
 
-The ``$bitNot`` aggregation stage performs a bitwise NOT operation on the given
-argument. ``$bitNot`` only takes one argument.
+The ``$bitNot`` aggregation operator performs a bitwise NOT operation on the given
+argument. You can use the ``$bitNot`` operator by preceding an 
+operand with a ``~`` character. ``$bitNot`` only takes one argument. The 
+following example shows how to create a ``$bitNot`` stage using LINQ:
 
 .. code-block:: csharp
 
-   var queryable = collection.AsQueryable()
+   var queryable = collection
       .Select(i => ~i.Count);
 
 The result of the preceding example contains the following values:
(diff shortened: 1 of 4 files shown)


source/fundamentals/linq.txt
```diff
@@ -587,6 +588,155 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
...
+$bitAnd
++++++++
+
+The ``$bitAnd`` aggregation stage performs a bitwise AND operation on the given
+arguments. The following example shows how to generate a ``$bitAnd`` stage using
+LINQ:
+
+.. code-block:: csharp
+
+    var query = queryableCollection.AsQueryable()
```
rustagir, line 643: S: you access your collection and call `AsQueryable()` at that time, so i dont think its necessary to include here. 
Also, would it be possible to move these examples into a code file and use an include?

source/fundamentals/linq.txt
```diff
@@ -587,6 +588,155 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
...
+   For all bitwise operators, if the operand list contains a missing or undefined
+   value, the entire expression evaluates to ``null``.
+
+
+$bitAnd
++++++++
+
+The ``$bitAnd`` aggregation stage performs a bitwise AND operation on the given
+arguments. The following example shows how to generate a ``$bitAnd`` stage using
+LINQ:
```
rustagir, line 639: S: right now, it's not totally clear that by bitAnd you mean the **&** operator. Perhaps this paragraph could be preceded by a short sentence that says something like

> You can use the ``$bitAnd`` operator by connecting two or more clauses with a ``&`` character.

Applies to all sections

source/fundamentals/linq.txt
```diff
@@ -587,6 +588,155 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
...
+.. code-block:: csharp
+
+    var query = queryableCollection.AsQueryable()
+      .Where(i => i.Name == "eggs")
+      .Select(i => i.Price & i.Count);
+
+The preceding example finds the document where the ``Name`` field has the value
+``"eggs"``. It then performs a bitwise AND operation on the values of the 
+``Price`` and ``Count`` fields in this document. The result contains one value:
+``4``.
```
rustagir, line 650:
```suggestion
The preceding example retrieves the document in which the ``Name`` field has the value
``"eggs"``. It then performs a bitwise AND operation on the values of the 
``Price`` and ``Count`` fields in this document, returning a value of ``4``.
```

source/fundamentals/linq.txt
```diff
@@ -587,6 +588,155 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
...
 in the Atlas manual and select :guilabel:`C#` from the language dropdown.
 
+Bitwise Operators
+~~~~~~~~~~~~~~~~~
+
+This version of the {+driver-short+} supports the following bitwise operators in 
+the aggregation pipeline. You can use multiple bitwise operators within the same 
+stage. All operands must be of type ``int`` or ``long``. ``$bitAnd``, ``$bitOr``,
+and ``$bitXor`` all take two or more operands. ``$bitNot`` can only take one
+operand.
```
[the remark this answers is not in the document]
rustagir: In this para, you could also briefly describe what a bitwise operator is by linking https://en.wikipedia.org/wiki/Bitwise_operation

source/fundamentals/linq.txt
```diff
@@ -587,6 +588,155 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
...
+The result of the preceding example contains the following values:
+
+.. code-block:: json
+
+   1
+   0
+   4
+   0
+   4
+   null
```
rustagir, line 669: S: you could combine these into an `io-code-block`

rustagir (changes requested):
Leaving a first batch of comments so that you can propagate these changes to the other sections. Please lmk if you have questions baout any of my feedback!

source/fundamentals/linq.txt
mayaraman19: thanks for catching this!

source/fundamentals/linq.txt
mayaraman19: added the suggested sentence after the first one since it flowed better that way in my opinion, but feel free to push back

source/fundamentals/linq.txt
```diff
@@ -587,6 +588,162 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
 {+driver-short+}, see :atlas:`Run Vector Search Queries </atlas-vector-search/vector-search-stage/>`
 in the Atlas manual and select :guilabel:`C#` from the language dropdown.
 
+Bitwise Operators
+~~~~~~~~~~~~~~~~~
+
+The following section describes the `bitwise operators <https://en.wikipedia.org/wiki/Bitwise_operation>`__
```
rustagir, line 594: S: use wiki role
```suggestion
The following section describes the :wikipedia:`bitwise operators </Bitwise_operation>`
```
Full list: https://github.com/mongodb/snooty-parser/blob/5dab215d105fa01976c6521c356323725e137a9d/snooty/rstspec-consolidated.toml

source/fundamentals/linq.txt
```diff
@@ -587,6 +588,162 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
...
+~~~~~~~~~~~~~~~~~
+
+The following section describes the `bitwise operators <https://en.wikipedia.org/wiki/Bitwise_operation>`__ 
+supported by the {+driver-short+} that you can use in an aggregation pipeline.
+You can use multiple bitwise operators in the same 
+stage. The following guidelines apply when using bitwise operators:
+
+- All operands must be of type ``int`` or ``long``.
+
+- ``$bitAnd``, ``$bitOr``, and ``$bitXor`` all take two or more operands. ``$bitNot`` can only take one operand.
```
rustagir, line 601:
```suggestion
- ``$bitAnd``, ``$bitOr``, and ``$bitXor`` take two or more operands. ``$bitNot`` takes one operand.
```

source/fundamentals/linq.txt
```diff
@@ -587,6 +588,162 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
...
+
+- The bitwise operation will be evaluated from left to right.
```
rustagir, line 603:
```suggestion
- Bitwise operations are evaluated from left to right.
```

source/fundamentals/linq.txt
```diff
@@ -587,6 +588,162 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
...
+   If the operand list you pass to any bitwise operator contains a missing or 
+   undefined value, the entire expression evaluates to ``null``.
+
+$bitAnd
++++++++
+
+The ``$bitAnd`` aggregation operator performs a bitwise AND operation on the given
+arguments. You can use the ``$bitAnd`` operator by connecting two or more 
+clauses with a ``&`` character. The following example shows how to create a 
+``$bitAnd`` stage using LINQ:
```
rustagir, line 637:
```suggestion
``$bitAnd`` stage by using LINQ: 
```

source/fundamentals/linq.txt
```diff
@@ -587,6 +588,162 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
...
+
+.. literalinclude:: /includes/fundamentals/code-examples/linq.cs
+   :language: csharp
+   :dedent:
+   :start-after: start-bitAnd-example
+   :end-before: end-bitAnd-example
+
+The preceding example retrieves the document in which the ``Name`` field has the 
+value ``"eggs"``. It then performs a bitwise AND operation on the values of the 
+``Price`` and ``Count`` fields in this document, returning a value of ``4``.
```
rustagir, line 647: S; move this to before the code example and only keep the result after the example
eg. 
The following code example retrieves the document, then performs a bitwise AND... : 

code

The preceding code returns ``4``, the result of the AND operation on the values ... and ...

rustagir: applies to all sections

rustagir (changes requested):
a few more small suggestions!

source/fundamentals/linq.txt
```diff
@@ -587,6 +588,169 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
 {+driver-short+}, see :atlas:`Run Vector Search Queries </atlas-vector-search/vector-search-stage/>`
 in the Atlas manual and select :guilabel:`C#` from the language dropdown.
 
+Bitwise Operators
+~~~~~~~~~~~~~~~~~
+
+The following section describes the :wikipedia:`bitwise operators </Bitwise_operation>`
```
rustagir, line 594:
```suggestion
This section describes the :wikipedia:`bitwise operators <Bitwise_operation>`
```

rustagir: got the wiki syntax wrong when i first suggested, sry

source/fundamentals/linq.txt
```diff
@@ -587,6 +588,169 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
...
+You can use multiple bitwise operators in the same 
+stage. The following guidelines apply when using bitwise operators:
+
+- All operands must be of type ``int`` or ``long``.
+
+- ``$bitAnd``, ``$bitOr``, and ``$bitXor`` take two or more operands. ``$bitNot`` takes one operand.
+
+- Bitwise operations are evaluated from left to right.
+
+The examples for this section use the following documents in a collection called
```
rustagir, line 605:
```suggestion
The examples in this section use the following documents in a collection called
```

source/fundamentals/linq.txt
```diff
@@ -587,6 +588,169 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
...
+   :language: csharp
+   :dedent:
+   :start-after: start-bitAnd-example
+   :end-before: end-bitAnd-example
+
+The preceding code returns ``4``, the result of the AND operation on the values
+of the ``Price`` field (``5``) and the ``Count`` field (``12``).
+
+The following example performs the same bitwise AND operation, but on all 
+documents in the collection:
```
rustagir, line 653:
```suggestion
The following example performs the same bitwise AND operation as in the preceding code on all 
documents in the collection:
```

rustagir (approved):
good job!! just a few nits

source/includes/fundamentals/code-examples/linq.cs
```diff
@@ -63,3 +63,56 @@ public class Review
...
+var query = queryableCollection
+    .Where(i => i.Name == "eggs")
+    .Select(i => i.Price & i.Count);
+
+// end-bitAnd-example
+
+// start-bitAnd-collection-example
+
+var query = queryableCollection
+    .Select(i => i.Price & i.Count);
```
rstam, line 93: This is an odd example. There's no real reason to bit-and a `Price` and a `Count`.

A better example would involve flags or something.

source/fundamentals/linq.txt
```diff
@@ -587,6 +588,169 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
...
+.. literalinclude:: /includes/fundamentals/code-examples/linq.cs
+   :language: csharp
+   :dedent:
+   :start-after: start-ingredient-model
+   :end-before: end-ingredient-model
+
+.. note:: Missing or Undefined Operands
+
+   If the operand list you pass to any bitwise operator contains a missing or 
+   undefined value, the entire expression evaluates to ``null``.
```
rstam, line 629: This is true server-side, but in C# this isn't true unless you are using "nullable" integers.

In these examples if either `Price` or `Count` is `null` the query will fail client-side while deserializing the results.

source/fundamentals/linq.txt
```diff
@@ -587,6 +588,169 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
...
+      1
+      0
+      4
+      0
+      4
+      null
+
+The ``null`` result comes from the document where the ``Name`` field
+has the value of ``"cheese"``. This document is missing a ``Count`` field, so
+the expression evaluates to ``null``.
```
rstam, line 677: Have you tried to execute this?

I think it will throw an exception because the `null` can't be deserialized to an `int` client-side.

source/fundamentals/linq.txt
```diff
@@ -587,6 +588,169 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
...
+``ingredients``:
+
+.. code-block:: json
+
+   { "_id": 1, "name": "watermelon", "price": 5, "count": 1 },
+   { "_id": 2, "name": "onions", "price": 2, "count": 4 },
+   { "_id": 3, "name": "eggs", "price": 5, "count": 12 },
+   { "_id": 4, "name": "potatoes", "price": 3, "count": 0 },
+   { "_id": 5, "name": "pasta", "price": 4, "count": 100 },
+   { "_id": 6, "name": "cheese", "price": 4 }
```
rstam, line 615: This final document can't have been created using the `Ingredient` POCO because `Count` is an `int` and can't be `null`.

Also, the element names in these sample documents don't match the field names in the POCO (the C# properties start with a capital letter).

source/fundamentals/linq.txt
```diff
@@ -587,6 +588,169 @@ in the Atlas manual. For more examples about running Atlas Vector Search queries
...
+   .. output:: 
+      :language: json
+      :visible: false
+
+      -2
+      -5
+      -13
+      -5
+      -101
+      null
```
rstam, line 726: `null` is not a possible result. It should throw an exception when deserializing the results.

rstam (changes requested):

source/includes/fundamentals/code-examples/linq.cs
mayaraman19: Changed the example!

source/fundamentals/linq.txt
mayaraman19: All examples in this PR have been executed.

source/fundamentals/linq.txt
mayaraman19: The data types used in the example have been changed to be nullable.

As for the second point, for these examples we use a convention pack (see the bottom of the [Overview section](https://www.mongodb.com/docs/drivers/csharp/current/fundamentals/linq/) on this page) to convert snake case field names into Pascal case.

source/fundamentals/linq.txt
mayaraman19: Clarified the note, please take a look!

rstam (approved):
LGTM

source/fundamentals/linq.txt
rstam: It's hard to tell from looking at just this PR that there is a convention somewhere else that is mapping the property names. That's confusing.

PR #264, opened 2024-10-17
GH action security updates

mongoKart:
# Pull Request Info


.github/workflows/add-netlify-links.yml
```diff
@@ -1,55 +1,75 @@
...
+      issues: write
+      contents: write
+      pull-requests: write
+      repository-projects: write
+    steps:
+      - uses: actions/checkout@v4
+      - name: Get Changed Files
+        id: changed-files
+        # pin to a specific commit to ensure stability
+        uses: tj-actions/changed-files@c65cd883420fd2eb864698a825fc4162dd94482c
```
MongoCaleb, line 19: Specifying the version (v44) ensures stability, too, and I don't usually see people referencing a specific commit id. I'd revert this.

.github/workflows/add-netlify-links.yml
```diff
@@ -1,55 +1,75 @@
...
+        with:
+          separator: ","
+          files: source/**
+      - name: Build Netlify Links for Changed Pages
+        id: build_page_links
+        env:
+          CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
+        run: |
+          # Function to validate file paths
+          validate_file_path() {
```
MongoCaleb, line 29: When I checked the source at tj-actions/changed-files, I _think_ I found that this check is already there. In other words, what it outputs are only truly changed files. It makes sense, since GitHub can't report on a changed file unless it is a valid file. 
This check might be a good safety net, but also complicates the steps and may be a source of issues if something else changes.

.github/workflows/add-netlify-links.yml
```diff
@@ -1,55 +1,75 @@
...
+          }
+
+          new_links=""
+          base_link='https://deploy-preview-${{ github.event.number }}--mongodb-docs-csharp.netlify.app'
+          files=$(echo "$CHANGED_FILES" | tr "," "\n")
+          for file in $files; do
+            echo "processing ${file}"
+            
+            # Validate file path and skip if invalid 
+            validate_file_path "$file"
```
MongoCaleb, line 45: related to my comment above.

.github/workflows/add-netlify-links.yml
```diff
@@ -1,55 +1,75 @@
 name: Add Netlify Links To Changed Pages
 on:
-    workflow_call:
+  workflow_call:
+  pull_request_target:
```
MongoCaleb, line 4: Is this needed? I think it's just if you want to specify a particular branch, which we don't want in this case.

MongoCaleb (commented):
If I'm seeing this correctly, the two main changes are 1) adding the env rather than using a local variable, and 2) adding a validation function. If there are others that I didn't comment on, please let me know!

.github/workflows/add-netlify-links.yml
mongoKart: This was new to me too. I found it in the [GitHub docs on hardening Actions.](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions). I think it's an extra layer of security against 1) unknown/unverified maintainers, 2) bad actors who gain access, and 3) obfuscated malicious code that slips into a version.

That said, I understand that it's less convenient. I could go either way but wanted to share why I changed it.

.github/workflows/add-netlify-links.yml
mongoKart: As long as this is an action whose code we don't monitor, it seems like there's always a risk of validation code being changed or removed. If we don't think the convenience tradeoff is worth it, I can remove it.

.github/workflows/add-netlify-links.yml
mongoKart: Which part do you mean? I removed `pull_request_target` this morning after Nic let me know about the vulnerability, just to close the door on it in the meantime. This PR adds it back.

.github/workflows/add-netlify-links.yml
MongoCaleb: Well that makes sense. According to All Things Git Holy, a tag is pinned to a specific commit and is immutable, but through some wrangling, it is possible for the repo owner to change it (https://stackoverflow.com/questions/21459540/add-new-commit-to-the-existing-git-tag). I suppose another approach would be to fork the repo and make it our own...but that's probably overkill. 
In short, keep it or revert it, and I'm happy. :)

.github/workflows/add-netlify-links.yml
MongoCaleb: Fair enough. I'm perhaps too trusting.

.github/workflows/add-netlify-links.yml
MongoCaleb: I see, OK.

MongoCaleb (approved):
LGTM

PR #365, opened 2024-12-09
DOCSP-41170 - Clarify string ID generation

mongoKart:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-41170

### Staging Links

<li><a href=https://deploy-preview-365--mongodb-docs-csharp.netlify.app/fundamentals/serialization/poco>fundamentals/serialization/poco</a></li>

## Self-Review Checklist


source/fundamentals/serialization/poco.txt
```diff
@@ -382,50 +382,86 @@ Specify an ID Generator
...
-
-The {+driver-short+} includes the following ``IIdGenerator`` types:
+The following table describes the generators available in the {+driver-short+}:
 
 .. list-table::
    :header-rows: 1
-   :widths: 10 10
+   :widths: 25 75
 
    * - ``Id`` Field Data Type
```
jordan-smith721, line 394: I can't immediately see why, but this is rendering without a space after "Id". Maybe because of the monospace?

source/fundamentals/serialization/poco.txt
```diff
@@ -382,50 +382,86 @@ Specify an ID Generator
...
+          :copyable: true
+
+          public class House
+          {
+             public Guid Id { get; set; }
+          } 
+
+   * - ``ObjectId``
+     - The driver automatically uses the ``ObjectIdGenerator`` type for ``Id`` fields with
+       the ``ObjectId`` data type, like the one in the following code example:
```
jordan-smith721, line 425: https://www.mongodb.com/docs/meta/style-guide/terminology/alphabetical-terms/#std-term-like
```suggestion
       the ``ObjectId`` data type, such as the one in the following code example:
```

mongoKart: boo this busted rule

source/fundamentals/serialization/poco.txt
mongoKart: i couldn't get this to look right so i just removed the monospace. seems like that actually matches the [style guide](https://www.mongodb.com/docs/meta/style-guide/style/titles-and-headings/#style-and-structure) too

rishitb-mongodb:
Could/Should we mention the example mentioned in the ticket as well that would allow users to define the generator in the attribute itself `[BsonId(IdGenerator = typeof(StringObjectIdGenerator))]`

mongoKart:
> Could/Should we mention the example mentioned in the ticket as well that would allow users to define the generator in the attribute itself `[BsonId(IdGenerator = typeof(StringObjectIdGenerator))]`

For my understanding: What is the use case for this attribute? Do you need to do this _in addition to_ the `[BsonRepresentation]` attribute in order to serialize to a string? Or can you use this attribute _instead of_ the other one?

rishitb-mongodb:
I'm actually not sure. Maybe @rstam can help. What's the difference between using `[BsonRepresentation(BsonType.ObjectId)]` on a string ID field as opposed to the attribute `[BsonId(IdGenerator = typeof(StringObjectIdGenerator))]`? If they are the same then we can ignore adding the alternative way, if they are different then its worth mentioning both and the difference between them.

rstam (approved):
LGTM

rstam:
>What's the difference between using [BsonRepresentation(BsonType.ObjectId)] on a string ID field as opposed to the attribute [BsonId(IdGenerator = typeof(StringObjectIdGenerator))]? 

They are not the same. Using BsonId with StringObjectIdGenerator without configuring the representation would store the ObjectIds in the database as strings.

BsonRepresentation(BsonType.ObjectId) stores the strings as ObjectIds in the database.

They are used together. Though unless you have changed the default conventions there is a convention that automatically sets the id generator for a string stored as an ObjectId to the StringObjectIdGenerator .

rishitb-mongodb:
Thanks Robert. @mongoKart I think its worth adding in a callout then in case a user wants to store the ID as a string along with representing it as a string in their POCO.

mongoKart:
@rishitb-mongodb @rstam Could you take another look?

rishitb-mongodb (approved):

source/fundamentals/serialization/poco.txt
```diff
@@ -261,6 +261,167 @@ The following code sample maps the ``Identifier`` property to the
...
+Specify an ID Generator
+``````````````````````
+
+Every document in a MongoDB collection must have a unique ID. When you serialize an object
+to a collection, if the ``Id`` property contains the default
+value for its data type (usually ``null``), the {+driver-short+} doesn't serialize the
+default value. Instead, the driver generates a unique ID value for the property.
+
+When you apply the ``[BsonId]`` attribute to a property, you can pass the ``IdGenerator``
+argument to specify the type of ID value the driver generates for the corresponding
```
rstam, line 273: Not "the type of ID value"

The type is already determined by the type of the property.

What are you specifying is which generator to use to generate an Id value when needed.

source/fundamentals/serialization/poco.txt
```diff
@@ -261,6 +261,167 @@ The following code sample maps the ``Identifier`` property to the
...
+       .. code-block:: csharp
+          :copyable: true
+
+          public class House
+          {
+             public ObjectId Id { get; set; }
+          } 
+
+   * - ``string``
+     - To serialize the ``Id`` property as a ``string``, apply the
```
rstam, line 323: Applying the `[BsonId]` attribute is NOT what causes this property to be serialized as a string.

All strings are serialized as strings UNLESS the `[BsonElement]` attribute specifies a different representation.

The ONLY thing this `[BsonId]` attribute is specifying is which id generator to use (beside identifying this as the ID field of course).

source/fundamentals/serialization/poco.txt
```diff
@@ -261,6 +261,167 @@ The following code sample maps the ``Identifier`` property to the
...
+          public class House
+          {
+             [BsonId(IdGenerator = typeof(StringObjectIdGenerator))]
+             public string Id { get; set; }
+          }
+         
+       To serialize the ``Id`` property as an ``ObjectId``, apply the
+       ``[BsonRepresentation(BsonType.ObjectId)]`` attribute to the property, as shown
+         in the following code example. In this case, if the property contains the value
+         ``null``, the driver uses the ``ObjectIdGenerator`` type to generate a unique
```
rstam, line 341: Actually it uses the `StringObjecIdGenerator`.

The `ObjectIdGenerator` can only be used with properties of type `ObjectId`.

source/fundamentals/serialization/poco.txt
```diff
@@ -261,6 +261,167 @@ The following code sample maps the ``Identifier`` property to the
...
+   
+            public class House
+            {
+               [BsonRepresentation(BsonType.ObjectId)]
+               public string Id { get; set; }
+            }
+       
+       If you apply both the ``[BsonId(IdGenerator = typeof(StringObjectIdGenerator))]``
+       and ``[BsonRepresentation(BsonType.ObjectId)]`` attributes to the ``Id`` property,
+       the driver generates an ``ObjectId`` value for the field.
```
rstam, line 355: Not sure this is worded the best way.

If you apply both you have simply done TWO different things:

1. Configured the representation of the ID in the database
2. Configured the id generator to use for missing ID values

rstam (changes requested):

source/fundamentals/serialization/poco.txt
mongoKart: After looking at the code for the Generator classes, I see that I've been conflating the ID generation and serialization steps.

source/fundamentals/serialization/poco.txt
```diff
@@ -247,7 +248,7 @@ The following code sample maps the ``Identifier`` property to the
        public string Identifier { get; set; }
    }
 
-.. warning:: Multiple Id Fields
+.. warning:: Multiple ID Fields
```
mongoKart, line 251: Changed most occurrences of 'Id' to 'ID' since you can use any name for an ID property

rstam: But... the C#/.Net naming convention is to use `Id`.

source/fundamentals/serialization/poco.txt
```diff
@@ -261,6 +262,166 @@ The following code sample maps the ``Identifier`` property to the
    The ``_id`` field mapping logic described in this section only applies to the
    root document and does not apply to nested documents.
 
+Specify an ID Generator
+```````````````````````
+
+Every document in a MongoDB collection must have a unique ID. When you serialize an object
+to a collection, if its ID property contains the default
+value for its data type (usually ``null``), the {+driver-short+} doesn't serialize the
+default value. Instead, the driver generates a unique ID value and assigns it to the
```
rstam, line 271: The C# driver ***TRIES*** to generate a unique Id value but can onlly do so in cases where it knows what Id generator to use.

rstam (approved):
I have two minor comments but if you want to push as is LGTM.

source/fundamentals/serialization/poco.txt
mongoKart: After talking to @jordan-smith721, keeping as 'ID' to comply with our Style Guide

jordan-smith721 (approved):

PR #373, opened 2024-12-12
DOCSP-40066: Support SelectMany inside Project/Select

mayaraman19:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-40066>

### Staging Links
- https://deploy-preview-373--mongodb-docs-csharp.netlify.app/fundamentals/linq/#nested-statements

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/linq.txt
+++ b/source/fundamentals/linq.txt
@@ -324,8 +324,11 @@ Nested Statements
 +++++++++++++++++
 
 You can chain or nest ``Select`` and ``SelectMany`` statements to unwind nested
-arrays. Consider a collection that holds documents with a ``restaurants`` field.
-This field contains an array of documents that represent the ``Restaurants`` class:
+arrays. Consider a collection that contains documents with a new schema. These
+documents contain a ``restaurants`` field, which holds an array of documents 
+that each represent the ``Restaurant`` class. These documents each have
+a ``grades`` field that holds an array of documents that each represent
+the ``Grade`` class.
 
 .. code-block:: json
 
@@ -388,6 +391,10 @@ in the preceding document.
             "grade" : "A", 
             "score" : 13.0 
          }, 
+         { "date" : ISODate("..."),
+           "grade" : ...,
+           "score" : ...
+         },
          ...
       ]
 
--- a/source/fundamentals/linq.txt
+++ b/source/fundamentals/linq.txt
@@ -368,8 +368,8 @@ this collection:
 
 You can nest ``SelectMany`` statements within ``SelectMany`` or ``Select``
 statements. The following example nests a ``SelectMany`` statement within a
-``Select`` statement to retrieve an array of ``Grade`` documents from all restaurants
-in each document in the collection.
+``Select`` statement to retrieve an array from each document in the collection.
+Each array holds all grade objects from all restaurants in each document.
 
 .. io-code-block::
    :copyable: true
--- a/source/fundamentals/linq.txt
+++ b/source/fundamentals/linq.txt
@@ -326,7 +326,7 @@ Nested Statements
 You can chain or nest ``Select`` and ``SelectMany`` statements to unwind nested
 arrays. Consider a collection that contains documents with a **new** schema. These
 documents contain a ``restaurants`` field, which holds an array of documents 
-that each represent the ``Restaurant`` class. These documents each have
+that each represent the ``Restaurant`` class. The documents within the array each have
 a ``grades`` field that holds an array of documents that each represent
 the ``Grade`` class. The following code is an example of a single document in
 this collection:
--- a/source/fundamentals/linq.txt
+++ b/source/fundamentals/linq.txt
@@ -328,7 +328,7 @@ arrays. Consider a collection that contains documents with a **new** schema. The
 documents contain a ``restaurants`` field, which holds an array of documents 
 that each represent the ``Restaurant`` class. These documents each have
 a ``grades`` field that holds an array of documents that each represent
-the ``Grade`` class.
+the ``Grade`` class. The following code models the document in this example:
 
 .. code-block:: json
 
@@ -368,7 +368,7 @@ the ``Grade`` class.
 You can nest ``SelectMany`` statements within ``SelectMany`` or ``Select``
 statements. The following example nests a ``SelectMany`` statement within a
 ``Select`` statement to retrieve an array of ``Grade`` documents from all restaurants
-in the preceding document.
+in each document in the collection.
 
 .. io-code-block::
    :copyable: true
@@ -382,6 +382,7 @@ in the preceding document.
       :visible: false
       :language: json
 
+      // output for first document in collection
       [
          { "date" : ISODate("2014-11-24T00:00:00Z"), 
             "grade" : "Z", 
@@ -395,6 +396,10 @@ in the preceding document.
            "grade" : ...,
            "score" : ...
          },
+         ...
+      ], 
+      // output for second document in collection
+      [
          ...
       ]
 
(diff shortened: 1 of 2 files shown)


source/fundamentals/linq.txt
```diff
@@ -320,6 +320,71 @@ following documents:
    { "date" : ISODate("2012-12-05T00:00:00Z"), "grade" : "A", "score" : 13 }
    { "date" : ISODate("2012-05-17T00:00:00Z"), "grade" : "A", "score" : 11 }
 
+Nested Statements
++++++++++++++++++
+
+You can chain or nest ``Select`` and ``SelectMany`` statements. The examples
+in this section use the ``Guidebook`` class, which contain an array of ``Restaurant``
+objects. The following code shows an example of a document modeled by the 
+``Guidebook`` class:
```
norareidy, line 329: Q: do you need to use another class? I think it makes this section a little long compared to the others; if you do need Guidebook, you could move it to "Overview"

source/fundamentals/linq.txt
```diff
@@ -320,6 +320,71 @@ following documents:
    { "date" : ISODate("2012-12-05T00:00:00Z"), "grade" : "A", "score" : 13 }
    { "date" : ISODate("2012-05-17T00:00:00Z"), "grade" : "A", "score" : 11 }
 
+Nested Statements
++++++++++++++++++
+
+You can chain or nest ``Select`` and ``SelectMany`` statements. The examples
+in this section use the ``Guidebook`` class, which contain an array of ``Restaurant``
```
norareidy, line 327: S:
```suggestion
in this section use the ``Guidebook`` class, which contains an array of ``Restaurant``
```

source/fundamentals/linq.txt
```diff
@@ -320,6 +320,71 @@ following documents:
...
+
+The following code models the ``Guidebook`` class:
+
+.. literalinclude:: /includes/fundamentals/code-examples/linq.cs
+   :language: csharp
+   :start-after: start-guidebook-model
+   :end-before: end-guidebook-model
+
+You can nest ``SelectMany`` statements within ``SelectMany`` or ``Select``
+statements. The following example nests a ``SelectMany`` statement within a
```
norareidy, line 360: Q: can you briefly explain why you'd want to nest these / what nesting the statements does? From my understanding (I'm not super familiar with c#), it would be something like:
```suggestion
statements to process nested document fields. The following example nests a ``SelectMany`` statement within a
```

source/fundamentals/linq.txt
```diff
@@ -320,6 +320,71 @@ following documents:
...
+``Select`` statement to retrieve an array of grades from all restaurants
+in a document modeled by the ``Guidebook`` class.
+
```
norareidy, line 363: I: since the examples on this page use the "sample_restaurants.restaurants" collection, which the "queryableCollection" variable represents, I think it gets confusing to introduce "Guidebooks". "Restaurants" kinda has to be the top-level class, since each instance represents a "restaurants" document. Is there a way to nest SelectMany within Select without introducing Guidebooks -- maybe selecting for restaurant documents with a certain "cuisine" value, then unwinding the grades arrays for just those documents?

norareidy (changes requested):
I left a couple questions, let me know if anything's unclear!

source/fundamentals/linq.txt
mayaraman19: While I agree that it is confusing to introduce "Guidebooks," unfortunately filtering by value requires `Where` instead of `Select`. The ideal example for this would involve nested arrays within a document. I tried to update the example to be a bit less involved, but I'm not sure if it's better or worse.

norareidy: ah okay! Then I think it's fine to introduce a new sample document

source/fundamentals/linq.txt
```diff
@@ -320,6 +320,77 @@ following documents:
...
+            ...
+         },
+         ...
+      ]
+   }
+
+You can nest ``SelectMany`` statements within ``SelectMany`` or ``Select``
+statements. The following example nests a ``SelectMany`` statement within a
+``Select`` statement to retrieve an array of ``Grade`` documents from all restaurants
+in the document above.
```
norareidy, line 368: S: "preceding" instead of "[above](https://www.mongodb.com/docs/meta/style-guide/terminology/alphabetical-terms/#above)":
```suggestion
in the preceding document.
```

norareidy (approved):
LGTM + small suggestion

source/includes/fundamentals/code-examples/linq.cs
```diff
@@ -81,6 +81,11 @@ public class Ingredient
 
 // end-ingredient-model
 
+// start-nested-SelectMany
+var query = queryableCollection
+    .Select(r => r.Restaurants.SelectMany(r => r.Grades));
```
papafe, line 86: I think here you wanted to use `.SelectMany`, instead of `.Select`

mayaraman19: I tried to make it a bit clearer that it's a document with a different schema.

As for the output, it's a little bit convoluted - the example is only printing out results for one document, and the way that we print on this page omits the outermost brackets since it prints each result separately. Therefore, while the output is technically the second, it looks like the first. I have edited the example a little bit to hopefully be a bit clearer.

papafe: Thanks, looks good!

papafe (approved):
LGTM

source/fundamentals/linq.txt
```diff
@@ -320,6 +320,90 @@ following documents:
    { "date" : ISODate("2012-12-05T00:00:00Z"), "grade" : "A", "score" : 13 }
    { "date" : ISODate("2012-05-17T00:00:00Z"), "grade" : "A", "score" : 11 }
 
+Nested Statements
++++++++++++++++++
+
+You can chain or nest ``Select`` and ``SelectMany`` statements to unwind nested
+arrays. Consider a collection that contains documents with a **new** schema. These
+documents contain a ``restaurants`` field, which holds an array of documents 
+that each represent the ``Restaurant`` class. The documents within the array each have
```
norareidy, line 329: S: since the classes model the documents, I think it's more accurate to say the document is "represented by" Restaurants, rather than represent Restaurants
```suggestion
represented by the ``Restaurant`` class. The documents within the array each have
```

source/fundamentals/linq.txt
```diff
@@ -320,6 +320,90 @@ following documents:
...
+a ``grades`` field that holds an array of documents that each represent
+the ``Grade`` class. The following code is an example of a single document in
```
norareidy, line 331: S: same suggestion as above
```suggestion
a ``grades`` field that holds an array of documents represented by
the ``Grade`` class. The following code is an example of a single document in
```

source/fundamentals/linq.txt
```diff
@@ -320,6 +320,90 @@ following documents:
...
+   .. input:: /includes/fundamentals/code-examples/linq.cs
+      :language: csharp
+      :start-after: start-nested-SelectMany
+      :end-before: end-nested-SelectMany
+
+   .. output::
+      :visible: false
+      :language: json
+
+      // output for first document in collection
```
norareidy, line 386: Q: should this be the first restaurant instead? since these are just the grades for Tov Kosher Kitchen, but the first guidebook document would have more restaurants
```suggestion
      // output for first restaurant
```

norareidy (approved):
LGTM with a few final thoughts!

source/fundamentals/linq.txt
mayaraman19: Nope, it should be first document - I will make the example/output more clear

PR #512, opened 2025-03-11
Set() stage builder

norareidy:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-33526

## Self-Review Checklist

- [ ] Is this free of any warnings or errors in the RST?
- [ ] Are the [facets and meta keywords](https://wiki.corp.mongodb.com/display/DE/Docs+Taxonomy) accurate?

### DIFF
--- a/source/fundamentals/builders.txt
+++ b/source/fundamentals/builders.txt
@@ -364,6 +364,43 @@ as a ``BsonDocument`` to the `AppendStage() method
 To learn more about the Aggregation Pipeline, see the
 :manual:`Aggregation Pipeline </core/aggregation-pipeline/>` server manual page.
 
+.. _csharp-builders-set:
+
+Add Fields to Collection Documents
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can add new fields to existing documents in a collection 
+by creating a ``$set`` stage in your aggregation pipeline. To
+create a ``$set`` stage, call the ``Set()`` method on a
+``PipelineStageDefinitionBuilder`` object. 
+
+.. tip::
+
+   To learn more about the $set stage, see :manual:`$set </reference/operator/aggregation/set/>`
+   in the {+mdb-server+} manual.
+
+This example builds an aggregation pipeline that performs the following
+actions:
+
+- Matches all documents that have a ``Name`` field value of ``"Daffodil"``
+- Adds a ``Color`` field to matching documents and sets its value to
+  ``"Yellow"``
+
+.. code-block:: csharp
+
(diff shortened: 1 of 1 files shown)


mongoKart (approved):
LGTM

source/fundamentals/builders.txt
```diff
@@ -364,6 +364,43 @@ as a ``BsonDocument`` to the `AppendStage() method
...
+  ``"Yellow"``
+
+.. code-block:: csharp
+
+   var matchFilter = Builders<Flower>.Filter.Eq(f => f.Name, "Daffodil");
+   var fields = Builders<Flower>.SetFields.Set("Color", "Yellow");
+
+   var pipeline = new EmptyPipelineDefinition<Flower>()
+      .Match(matchFilter)
+      .Set(fields);
```
rstam, line 396: Have you tried executing this?

Most likely it will throw an exception when reading the results from the server because the C# declaration for `Flower` does not have a `Color` property.

rstam (changes requested):

source/fundamentals/builders.txt
norareidy: Ah yep, I added `Color` to the `Flower` class while testing and just added it to the page. But does that defeat the purpose of `Set()`?

source/fundamentals/builders.txt
```diff
@@ -364,6 +364,43 @@ as a ``BsonDocument`` to the `AppendStage() method
 To learn more about the Aggregation Pipeline, see the
 :manual:`Aggregation Pipeline </core/aggregation-pipeline/>` server manual page.
 
+.. _csharp-builders-set:
+
+Add Fields to Collection Documents
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can add new fields to existing documents in a collection
```
rstam, line 372: This might be a bit misleading because the documents in the collection are NOT modified.

The new fields are only added to the documents going through the *pipeline*, the original documents that are input to the pipeline are unchanged.

source/fundamentals/builders.txt
rstam: Good point. It looks like the API for `Set` is defective. Let me discuss with the team.

mongoKart:
@norareidy I added a C# example to the $set Server page; do you think we can close this PR?

https://www.mongodb.com/docs/upcoming/reference/operator/aggregation/set/#examples

norareidy:
> @norareidy I added a C# example to the $set Server page; do you think we can close this PR?
> 
> https://www.mongodb.com/docs/upcoming/reference/operator/aggregation/set/#examples

If Set() is now working as expected I can just merge this PR; @rstam are there any updates on that?

